PR #1, opened 2018-06-07
2.1 pre-release docs

jamiltz:
So far this PR contains docs for the message endpoint API in Swift. Before doing the same for other languages/sdks, I suggest that we do a review of this PR first.

Preview http://docs-build.sc.couchbase.com/antora/site/couchbase-lite/2.1/swift.html#message-endpoint

Docs for other features (repl checkpoint and encryption) are already on master.


modules/ROOT/pages/swift.adoc
```diff
@@ -850,6 +850,12 @@ the result.
 Couchbase Mobile 2.0 uses a new replication protocol based on
 WebSockets.
 
+The replicator is designed to send documents from a source to a target database. The target can be one of the following:
+
+- `URLEndpoint`: to replicate data between a local Couchbase Lite database and remote Sync Gateway database using the WebSocket protocol ([start here]).
+- `DatabaseEndpoint`: to replicate data between two local Couchbase Lite databases. This target can be used to create a local back-up of a database ([database replicas])
+- `MessageEndpoint`: allows a local database to replicate with another Couchbase Lite database via a custom transportation protocol such iOS Multipeer Connectivity, Android WiFi Direct, Android NearByConnection, socket based transportation etc. (see xref:#message-endpoint[Message Endpoint])
+
```
borrrden: Call out EE only for the above two?

jamiltz: Indeed, I've pushed a commit to point this out for encryption and the message endpoint API. ([preview](http://docs-build.sc.couchbase.com/antora/site/couchbase-lite/2.1/swift.html#message-endpoint))

jamiltz:
@borrrden I have updated the PR with Priya's extensive p2p review. The idea is to describe the steps to implement the Message Endpoint API with as much detail as possible. I think that 90% of the explanations will also apply to the C# SDK although class names may differ. As a first step, it would be good to have the equivalent code snippets in C# ([sample code file](https://github.com/couchbaselabs/couchbase-lite-docs/blob/pre-release/2.1/modules/ROOT/examples/swift/SampleCodeTest.swift#L703)). If some class names/properties or even a paragraph doesn't make any sense feel free to bring it up. I think it's going to take a couple more passes to make this section as easy to read through as possible.

Preview: http://docs-build.sc.couchbase.com/antora/site/couchbase-lite/2.1/swift.html#message-endpoint

borrrden:
I've already pretty much done that [here](https://drive.google.com/file/d/1hR3Hsi_m6IqUOKaT65qwq3issW7d7QJ0/view?usp=sharing) FYI.

jamiltz:
That document is great and goes into details that the content on this PR doesn’t. So I will take the bits missing and add them here.

> I'm worried about putting so much code here that will be unsupported...

What do you mean by unsupported?

I think we need to make it very clear to the reader that replication won’t actually work until the communication framework code is added where required. But at least, having the Message Endpoint API classes in place is probably a good place to start.

borrrden:
Unsupported means that even though these code snippets are in the docs, these are not supported tested snippets of code and we won't be fielding support requests for it.  It is merely to help people get started on making their own and to help them be familiar with what the various interfaces are for.

jamiltz:
@borrrden I have added a note to warn users that replication won't work until the comms framework code is added.

> Unsupported means that even though these code snippets are in the docs, these are not supported tested snippets of code

That's true but they are merely there to describe the API. The other way we could document those steps would be to have paragraphs of text with links to the API ref methods.

jamiltz:
Closing this PR to avoid being notified of every commit that gets pushed. I will re-open it when it's ready for final review before merging to master.

PR #3, opened 2018-06-15
encryption specification

jamiltz:


modules/ROOT/pages/_partials/database-encryption.adoc
```diff
@@ -1,5 +1,7 @@
 Couchbase Lite 2.1 introduces database encryption, this functionality is exclusively available in the Enterprise Edition.
 
-The xref:#database-upgrade[automatic database upgrade] functionality is *not supported* for encrypted databases. Furthermore, databases encrypted with one SDK (i.e Swift, C#, Java or Objective-C) *cannot* be opened with another SDK.
+If you're migrating an application from Couchbase Lite 1.x to 2.x, note that the xref:#database-upgrade[automatic database upgrade] functionality is *not supported* for encrypted databases. Furthermore, databases encrypted with one SDK (i.e Swift, C#, Java or Objective-C) *cannot* be opened with another SDK.
```
rajagp: Can you clarify what you meant by "databases encrypted with one SDK ..". Did you mean some one creates CBLite db on Android and tries to use that in an iOS app? Maybe clarify that with an example then ?

jamiltz: ok I've updated the text (see commit below). TBH I don't even know if that is still true. That documentation came in during dev builds but maybe it's not a limitation anymore.

PR #5, opened 2018-06-19
ios deployment considerations

jamiltz:
https://issues.couchbase.com/browse/DOC-3734


modules/ROOT/pages/_partials/app-store-script.adoc
```diff
@@ -0,0 +1,15 @@
+=== Submission to the App Store
+
+The following does not apply if you have installed Couchbase Lite with Carthage or Cocoapods. With those package managers, Couchbase Lite is built with the application itself according to the architecture type of the target (emulator or device).
```
rajagp: I don't think that that would be the case for Carthage because you pull and build the binaries separate from the project . You should probably link folks to https://github.com/Carthage/Carthage/blob/5fd867c4895b4f59d70181dec169a1644f4430e3/README.md#if-youre-building-for-ios

jamiltz: Oh I see! carthage users will have to run `carthage copy-frameworks`. Ok will update.

modules/ROOT/pages/_partials/app-store-script.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+The Couchbase Lite framework available on the link:https://couchbase.com/downloads[downloads page] contains a build for both the simulator (x86_64) and iOS devices (ARM). Submissions to the App Store of a binary that contains the simulator architecture is not supported therefor you must remove it from **CouchbaseLite.framework** before App Store submission. The following steps describe how to set up a build phase in Xcode to do this automatically.
```
rajagp: Just occurred to me that we may have an issue even w/ Cocoapods since with Couchbase Lite EE version that's distributed via binaries only.  Workarounds specified here  may not quite work  since they update settings to be arm only ...https://github.com/CocoaPods/CocoaPods/issues/7111 . (@pasin )

jamiltz:
@rajagp I have update to include the link for Carthage.

Ok for Cocoapods. I suggest that we merge this PR (if it looks good for Frameworks/Carthage) and open a new one for Cocoapods when there's more info for that. Is that ok?

PR #7, opened 2018-06-25
add api-walkthrough project

jamiltz:
relates to https://github.com/couchbase/couchbase-lite-net/pull/1035


jamiltz:
@borrrden not sure if it still compiles. The couchbase-lite nuget reference may need updating.

borrrden:
If it doesn’t then we have the usual chicken egg problem because there is no newer public release.

jamiltz:
One way to get around this problem would be to `git clone` couchbase-lite-net in a **libs** directory and use a relative reference instead of nuget. That's what we're doing for android https://github.com/couchbaselabs/couchbase-lite-docs/blob/master/modules/ROOT/examples/java/code-snippets/app/build.gradle#L29-L30, following those instructions to set it up https://developer.couchbase.com/documentation/mobile/2.0/guides/couchbase-lite/tech-notes/index.html?language=java

borrrden:
That’s not a great alternative either as this is meant for users to build and look at. They will have a hard time building the whole project as it is far more complex than android.

jamiltz:
I guess it’s a trade off between the chicken and egg problem and ease of use for external users then.

jamiltz:
@borrrden can we point the api-walkthrough project to the pre-release version (currently DB1)? And bump it when the next release is available on the public nuget feed. And then on the master branch, point it to a Couchbase CI nuget feed which may require VPN access for the latest builds (if such a feed exists).

borrrden:
I thought it already was pointing to the 2.1 db. I’m not clear on what you want to do with the master branch. Are you thinking that it will pull the latest package like android snapshots?  That won’t really work unless we remove the feed with the external developer builds because external developer builds sort higher than internal ones (“db” vs “b”). There is not a good answer for this I think. By the way what features are being used that are not available to compile at the moment?

jamiltz:
Oh yeah it is pointing to 2.1. I'll go ahead and merge then.

> By the way what features are being used that are not available to compile at the moment?

None at the moment. But P2P might be one to document soon that isn't in DB1. In this scenario, the project on the master won't compile but users shouldn't try to use it anyway until a release/pre-released is publicly available. And I don't expect it to be used much externally anyway, the api-walkthrough is more of a point of reference for the docs/eng teams to be aligned.

PR #8, opened 2018-06-27
Extract version into asciidoc attribute

jamiltz:
@pasin could you check if the version that I have specified in this commit for 2.1 DB1 (java/swift/objc) is correct?


borrrden:
I think the version should be 2.1.0-DB001

jamiltz:
Thanks!

PR #9, opened 2018-07-17
update replication code example

jamiltz:
relates to https://issues.couchbase.com/browse/DOC-2842


jamiltz:
There has been confusion on how to setup a replication (as discussed in https://issues.couchbase.com/browse/CBSE-5304). TLDR, this PR updates the code snippets to use an instance variable and the class it belongs to. It should help readers into doing the right thing from the start.

@borrrden would it be possible to have the same for C#? I'll keep this PR open, let me know if you have any questions.

borrrden:
I'm slightly unclear as to what exactly needs to happen.  Could you put in a placeholder for the class?

jamiltz:
Placeholder commit added. The sample code for other languages is also on this PR.

borrrden:
Sorry for how slow that was, I was traveling and such.  I added an example.

jamiltz:
No problem!

PR #12, opened 2018-08-07
use incremental subs instead of explicit subs

mojavelinux:
By specifying the attributes substitution, you loose the special characters substitutions, which is crucial for not creating broken HTML.

You almost never want to specify subs explicitly unless you are going to specify all of them. Always start with incremental subs and then get more explicit if necessary.

See https://asciidoctor.org/docs/user-manual/#incremental-substitutions for details.


jamiltz:
Sounds good. Thanks!

mojavelinux:
We'll need to look out for this in sync gateway as well. I'll submit PRs when I come across them.

jamiltz:
Ok - I will do a follow up of all the merged PRs to cherry pick to other branches. And I'll update other instances of `"attributes"` to `attributes+` in the process.

PR #13, opened 2018-08-07
use source-language attribute instead of lang

mojavelinux:
The `lang` attribute is reserved for human language. Use the `source-language` attribute to specify the programming language instead.

Additionally, when you set the source-language attribute, you do not need to specify the language on the source block as it gets inherited automatically.

```
[source]
----
include::{examplesdir}/{snippetsdir}[tags=getting-started]
----
```

See https://asciidoctor.org/docs/user-manual/#applying-source-highlighting for details.


jamiltz:
Ok!

PR #14, opened 2018-08-07
clean up AsciiDoc

mojavelinux:
- simplify table delimiters
- remove unnecessary table attributes
- use list-style table rows
- change standalone images to block images
- replace setext titles with atx titles
- remove explicit number from ordered list items
- remove _attributes.adoc files (we're taking a different approach)


jamiltz:
Thanks!

PR #18, opened 2018-08-08
simplify xref syntax

mojavelinux:
- drop ROOT module when referring to another component (it's implied)
- use xref shorthand syntax to link within page

When referring to an anchor in the same page, the syntax `<<id,text>>` is preferred. That way, it's easy to spot an internal reference vs a local reference.


jamiltz:
Thanks!

PR #22, opened 2018-08-09
use tiles layout for samples page

mojavelinux:
- shift all heading levels up by one
- set the page role to tiles, which activates the tiles content layout
- disable sectids to skip toc creation


jamiltz:
Looks great. Thanks!

PR #23, opened 2018-08-10
resolves #11 add missing doctitles and add section titles where needed (master)

### LINKED ISSUE #11: Use level-0 section for page title
@jamiltz Can you update the AsciiDoc pages to use a level-0 section for the page (aka document) title?

```
= Compatibility Matrix
  
The table below summarizes the compatible versions of Couchbase Lite with Sync Gateway.
```

Both Antora and the UI design rely on the document starting with an H1 heading (which is what the page title becomes).

mojavelinux:
- shift all section levels so documents begin with a doctitle

### DIFF
--- a/modules/ROOT/pages/compatibility-matrix.adoc
+++ b/modules/ROOT/pages/compatibility-matrix.adoc
@@ -1,4 +1,4 @@
-== Compatibility Matrix
+= Compatibility Matrix
 
 The table below summarizes the compatible versions of Couchbase Lite with Sync Gateway.
 
--- a/modules/ROOT/pages/csharp.adoc
+++ b/modules/ROOT/pages/csharp.adoc
@@ -1,7 +1,8 @@
-:source-language: csharp
+= C#
 :idprefix:
 :idseparator: -
 :snippetsdir: csharp/Program.cs
+:source-language: csharp
 
 == Getting Started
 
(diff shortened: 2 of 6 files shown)


jamiltz:
I've tried building with the changes from this PR. The TOC has a lot more items because all the h3 are now h2s. I think that having that many items in the nav is actually nice, but perhaps with some indentation between h2s and h3s. That's another topic of discussion though. This PR would also affect the TOC on the [current dev portal](https://developer.couchbase.com/documentation/mobile/2.0/couchbase-lite/swift.html) (it looks for all h2s) so we might have to postpone making this headings change if that's ok. I had not foreseen this output change.

Before

image


After

image

mojavelinux:
The TOC on the new site does not include h3 headings. It's built primarily from h2 headings, as well as h1 in cases where they are used in the flow. So the problem here is not lack of indentation. It's that I didn't realize that h2 headings already used in these documents...so shifting all levels up is not the correct solution in this case. Instead, we just need to introduce the missing doctitle. We can't just sweep this structural problem under the rug. It has to be fixed.

I'll review the PR with a new solution.

mojavelinux:
@jamiltz okay, give it a try now. it should match the behavior of the old site with the addition of the doctitles.

jamiltz:
Yes it does. Thanks!

PR #31, opened 2018-08-15
optimize compatibility matrix table (master)

mojavelinux:
* use implicit header row (defined on a single row followed by a blank line)
* apply the header style to the sync gateway column header


jamiltz:
Looks much nicer 👍

PR #32, opened 2018-08-15
use proper AsciiDoc admonitions (master)

mojavelinux:
Note that Antora doesn't currently support a custom caption on an admonition block. A block title should be used instead.


jamiltz:
Ok.

PR #33, opened 2018-08-15
refactor source listings (master)

mojavelinux:
- consolidate include target to {snippet}
- drop redundant language declaration on source block
- rename tags attribute to tag
- use {source-language} instead of {lang}
- rename objc to objectivec


jamiltz:
Great clean-up! And will save us a lot of typing for future work.

PR #44, opened 2018-08-16
move nav into ROOT; drop internal refs (release/1.4)

mojavelinux:
Let's not add internal references to the nav as that area is reserved for pages. The internal refs show up in the "On This Page" on the right.


jamiltz:
Ok.

PR #46, opened 2018-08-16
change relative URLs to absolute URLs (release/1.4)

mojavelinux:
* change relative URLs to absolute
* drop redundant link: prefix
* add missing link targets


mojavelinux:
No problem. That was to be expected. Update as needed.

PR #55, opened 2018-08-17
fix swift snippets (release/2.0)

mojavelinux:
- use correct path to snippet
- fix misspelled closing tag


mojavelinux:
Preceded by #56.

PR #59, opened 2018-08-29
escape literal ++ in string

mojavelinux:
The character sequence ++ has very strong meaning in AsciiDoc. If it occurs more than once on a line, it pairs up to create a passthrough. There are two ways to escape it.

* Define the `pp` attribute and use it in place of the literal text `++` (the `pp` attribute will eventually be a built-in, but for now we have to define it)
* Place it inside an inline pass macro (e.g., `pass:[libc++]`)

I opted to go with the first solution in this PR.


jamiltz:
ok!

PR #67, opened 2019-02-15
Fix reference to Xcode in Android Studio chapter

PaoloneM:
Change to Android to match the context. Being cblite compatible with android framework only, reference to Android might be more suitable and not misleading than referring to Java


jamesnocentini:
@PaoloneM Thanks!

PR #70, opened 2019-03-09
Feature/doc 4793

Sandychuang8:
Adds a query change listener.

PR #74, opened 2019-03-18
DOC-4793

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4793

### DIFF
--- a/modules/ROOT/examples/csharp/Program.cs
+++ b/modules/ROOT/examples/csharp/Program.cs
@@ -320,19 +320,17 @@ namespace api_walkthrough
                     /* Update UI */
                 }
             });
+			
+			// Start live query.
+            query.Execute(); // <1>
 
-            try {
-                // Start live query.
-                query.Execute(); // <1>
-            } finally {
-                // tag::stop-live-query[]
-                query.RemoveChangeListener(token);
-                // end::stop-live-query[]
+            // tag::stop-live-query[]
+            query.RemoveChangeListener(token);
+            // end::stop-live-query[]
 
-                Console.WriteLine(query);
+            Console.WriteLine(query);
 
-                query.Dispose();
-            }
+            query.Dispose();
             // end::live-query[]
         }
 
--- a/modules/ROOT/examples/csharp/Program.cs
+++ b/modules/ROOT/examples/csharp/Program.cs
@@ -308,7 +308,7 @@ namespace api_walkthrough
             // tag::live-query[]
             var query = QueryBuilder
                 .Select(SelectResult.All())
-                .From(DataSource.Database(Db));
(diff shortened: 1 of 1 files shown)


jamesnocentini:
In the current form of the code snippet, `query.RemoveChangeListener(token);` is being called immediately after `query.Execute()` and this may be confusing to users who expect the code example to work as is.

@Sandychuang8 @borrrden is there a different syntax in C# to separate the live query setup and live query listener removal? I changed something here but I'm pretty sure that wouldn't compile.

borrrden:
A `using` statement is just sugar for the following:

```c#
Query q;
try {
    q = <whatever>
    q.DoStuff();
} finally {
    q.Dispose();
}
```

So if you want to separate them just remove the using statement all together and be sure to call `Dispose()` later.

jamesnocentini:
@borrrden ok thanks.

@Sandychuang8 Sorry I didn't notice that until the [previous PR](https://github.com/couchbase/docs-couchbase-lite/pull/70) was merged. Might it be possible to update it per Jim's suggestion? You can push a commit this PR. Thanks!

(in swift [live-query](https://github.com/couchbase/docs-couchbase-lite/blob/release/2.1/modules/ROOT/examples/swift/code-snippets/SampleCodeTest.swift#L226) and [stop-live-query](https://github.com/couchbase/docs-couchbase-lite/blob/release/2.1/modules/ROOT/examples/swift/code-snippets/SampleCodeTest.swift#L244))

Sandychuang8:
@jamiltz It's a good practice to put query into a `using` statement. I don't think we should modify what we have.

modules/ROOT/examples/csharp/Program.cs
```diff
@@ -306,30 +306,31 @@ private static void SelectAll()
...
+            var token = query.AddChangeListener((sender, args) =>
+            {
+                var allResult = args.Results.AllResults();
+                foreach (var result in allResult) {
+                    Console.WriteLine(result.Keys);
+                    /* Update UI */
+                }
+            });
+			
+			// Start live query.
```
jamesnocentini:
```suggestion
            // Start live query.
```

modules/ROOT/examples/csharp/Program.cs
```diff
@@ -306,30 +306,31 @@ private static void SelectAll()
             // end::query-select-all[]
 
             // tag::live-query[]
-            using (var query = QueryBuilder
+            var query = QueryBuilder
                 .Select(SelectResult.All())
-                .From(DataSource.Database(db))) {
+                .From(DataSource.Database(Db));
```
jamesnocentini:
```suggestion
                .From(DataSource.Database(db));
```

borrrden (approved):

PR #75, opened 2019-03-25
DOC-4809

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4809

### DIFF
--- a/modules/ROOT/pages/_partials/query/date-time-functions.adoc
+++ b/modules/ROOT/pages/_partials/query/date-time-functions.adoc
@@ -3,11 +3,15 @@ Couchbase Lite documents support a <<initializers,date type>> that internally st
 Couchbase Lite 2.5 adds the ability to run date comparisons in your Couchbase Lite queries.
 To do so, four functions have been added to the Query Builder API:
 
-* `Function.StringToMillis(Expression.Property("date_time"))`: The input to this will be a validly formatted ISO 8601 date time string.
+`Function.StringToMillis(Expression.Property("date_time"))`::
+The input to this will be a validly formatted ISO 8601 `date_time` string.
 The end result will be an expression (with a numeric content) that can be further input into the query builder.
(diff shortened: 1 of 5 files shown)


modules/ROOT/pages/_partials/query/date-time-functions.adoc
```diff
@@ -0,0 +1,13 @@
+Couchbase Lite documents support a <<initializers,date type>> that internally stores dates in ISO 8601 with the GMT/UTC timezone.
+
+Couchbase Lite 2.5 adds the ability to run date comparisons in your Couchbase Lite queries.
+To do so, four functions have been added to the Query Builder API:
+
```
eric-schneider, line 5: Here are some content edits, along with a different idea for presenting them if you're interested.

```
`Function.StringToMillis(Expression.Property("date_time"))`::
The input to this will be a validly formatted ISO 8601 `date_time` string.
The end result will be an expression (with a numeric content) that can be further input into the query builder.
`Function.StringToUTC(Expression.Property("date_time"))`::
The input to this will be a validly formatted ISO 8601 `date_time` string.
The end result will be an expression (with string content) that can be further input into the query builder.
`Function.MillisToString(Expression.Property("date_time"))`::
The input for this is a numeric value representing milliseconds since the Unix epoch.
The end result will be an expression (with string content representing the date and time as an ISO 8601 string in the device’s timezone) that can be further input into the query builder.
`Function.MillisToUTC(Expression.Property("date_time"))`::
The input for this is a numeric value representing milliseconds since the Unix epoch.
The end result will be an expression (with string content representing the date and time as a UTC ISO 8601 string) that can be further input into the query builder.
```

eric-schneider (commented):
Provided some edits and a different way of presenting the list of functions.

modules/ROOT/pages/_partials/query/date-time-functions.adoc
jamesnocentini: Yes! Applied in https://github.com/couchbase/docs-couchbase-lite/pull/75/commits/9e49857002fb3f559753c3b40a973c04807cad02

PR #82, opened 2019-04-06
Add example for ObjC

jayahariv:


modules/ROOT/examples/objc/code-snippets/SampleCodeTest.m
```diff
@@ -22,9 +22,9 @@
 #import <CoreML/CoreML.h>
 
 // tag::predictive-model[]
-// `tensorFlowModel` is a fake implementation
+// `coreMLModel` is a fake implementation
 // this would be the implementation of the ml model you have chosen
-@interface tensorFlowModel : NSObject
+@interface coreMLModel : NSObject
```
jamesnocentini: @jayahariv Naming the fake implementation `coreMLModel` might lead users to do more work than is needed. For CoreML models, iOS users should use the convenient [CoreMLPredictiveModel](https://docs-staging.couchbase.com/couchbase-lite/2.5/swift.html#integrate-a-model-with-coremlpredictivemodel-ios-only) API instead of implementing `CBLPredictiveModel`.

However, to be model agnostic, we could rename it to `MLModel`.

jamesnocentini: Confirmed with Priya we can rename it to `MLModel`. I'll do it for Swift.

jamesnocentini (changes requested):

modules/ROOT/examples/objc/code-snippets/SampleCodeTest.m
jayahariv: `MLModel` is already a taken interface name, `customMLModel`, or `aMLModel`?

jamesnocentini: Naming is indeed hard. Having thought about it more. It might be more explicit to include the word Framework in the name as well (to suggest the usage of a 3rd-party ML framework).

Suggest `anyMLFramework` ([698f82f](https://github.com/couchbase/docs-couchbase-lite/commit/7e8c7b6c5801bbad0ef99d81883870981698f82f)). @rajagp what do you think?

rajagp: 'anyMLModel' or 'myMLModel' is fine with me...I don't want to confuse model with the framework (framework leverages the model)

jamesnocentini: OK. Updated for Swift ([307bd49](https://github.com/couchbase/docs-couchbase-lite/commit/307bd49475de7b5c93dcfc5afbe422aa9830d68e)).

jayahariv: Updated for ObjC (6e0432537826229bf249e889779c8f39a92b9d73)

jamesnocentini (approved):

PR #83, opened 2019-04-09
DOC-5051

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5051

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -6,4 +6,4 @@
 * xref:csharp.adoc[C#]
 * xref:objc.adoc[Objective-C]
 * xref:samples.adoc[Samples]
-* xref:compatibility-matrix.adoc[Compatibility Matrix]
+* xref:compatibility-matrix.adoc[Compatibility]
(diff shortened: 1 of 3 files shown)


jamesnocentini:
Screenshot: 
image

jamesnocentini:
@rajagp If I recall correctly, CBL and CBS SDKs are aligned on the usage of interface stability terms so suggesting to reuse the same.

@RichardSmedley you already had a partial for this section of SDK docs so I went ahead and used that. It's the first time CBL docs pull content from the sdk repo so it would be good to get your thoughts on this approach.

rajagp:
@jamiltz  Yes, we are aligned on definition. Also, is there a place on server docs where term "Developer Preview" is defined (asking in context of predictive query stuff)

RichardSmedley (approved):
Interface stability partial import LGTM :+1:

jamesnocentini:
@rajagp After checking with the wider doc team it appears there isn't a page defining "Developer Preview" on docs.couchbase.com yet. Can discuss on Slack how we want to handle that for 2.5.

rajagp (approved):

PR #85, opened 2019-04-18
note about filtering out access removals

jamesnocentini:


modules/ROOT/pages/_partials/replication-filters.adoc
```diff
@@ -30,4 +30,9 @@ include::{snippet}[tag=replication-pull-filter,indent=0]
...
 <1> The callback should follow the semantics of a https://en.wikipedia.org/wiki/Pure_function[pure function].
 Otherwise, long running functions would considerably slow down the replicator.
-Furthermore, your callback should not make assumptions about what thread it is being called on.
\ No newline at end of file
+Furthermore, your callback should not make assumptions about what thread it is being called on.
+
+NOTE: Losing access to a document (via the Sync Function) also triggers the pull replication filter.
+Filtering out such an event would retain the document locally.
+As a result, there would be a local copy of the document disjointed from the one that resides on Couchbase Server.
+Further updates to the document stored on Couchbase Server would not be received in pull replications.
```
borrrden: Add in the second half of this, which is "but further edits could be potentially pushed which could result in either confusion as someone who has "lost access" would be pushing, or extra pusher work as it gets a 409 each time that it cannot ever resolve because of lack of pull access"

borrrden (changes requested):

modules/ROOT/pages/_partials/replication-filters.adoc
jamesnocentini: OK thanks, will apply that edit.

jamesnocentini:
```suggestion
Further updates to the document stored on Couchbase Server would not be received in pull replications and further local edits could be potentially pushed, which would result in 409 errors since access has been revoked.
```

PR #96, opened 2019-05-21
API references page attribute

jamesnocentini:
add `url-api-references` page attribute and replace throughout page

### DIFF
--- a/modules/ROOT/pages/csharp.adoc
+++ b/modules/ROOT/pages/csharp.adoc
@@ -5,6 +5,7 @@
 :source-language: csharp
 :blank-field: ____
 :url-issues-net: https://github.com/couchbase/couchbase-lite-net/issues
+:url-api-references: http://docs.couchbase.com/mobile/2.5.0/couchbase-lite-net
 
 == Getting Started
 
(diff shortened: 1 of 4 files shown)


simon-dew:
Out of interest, why do these docs contain links to the compiled html pages, rather than xrefs to the AsciiDoc source?

jamesnocentini:
Glad you asked. API refs (for [swift](https://docs.couchbase.com/mobile/2.5.0/couchbase-lite-swift/) for example) are generated with tools specific to each platform (Jazzy for Swift/ObjC, Sandcastle for C# and JavaDoc for Java). We could maintain an API ref in AsciiDoc itself, but, developers tend to prefer the format generated by the tools on the platform of their choice (e.g JavaDoc is very popular within the Java community). It's also much easier for us to generate API refs from a command line than to ensure correctness by hand (probably the biggest reason to justify that choice).

simon-dew:
Ah, understood!

simon-dew (approved):
Approved, when [conflicts](https://github.com/couchbase/docs-couchbase-lite/pull/96/conflicts) are resolved

PR #99, opened 2019-05-21
Update PinnedServerCertificate

Sandychuang8:


jamesnocentini:
@Sandychuang8 Looks good, I have one question about the `GetCertificateFromStore` method.

@borrrden @Sandychuang8 the docs will pull the code that is between tags ([here](https://github.com/couchbase/docs-couchbase-lite/blob/doc-5266/modules/ROOT/examples/csharp/Program.cs#L806-L817)) so it won't have the `GetCertificateFromStore` implementation which may leave users wondering what to do.
In Swift docs, we load it from the Bundle ([link](https://docs.couchbase.com/couchbase-lite/2.5/swift.html#certificate-pinning)) for the moment (I'm finding out if that is good or bad practice).
The choice for C# is:
1. do we prefer to document how to retrieve from the `X509Store` and let the reader figure how to get the certificate into that store in the first place.
2. do we prefer to load it from a filesystem path? if loading from a filesystem path is bad practice we should definitely go with option 1. By system path, I mean something similar to the [loading a pre-built database](https://docs.couchbase.com/couchbase-lite/2.5/csharp.html#loading-a-pre-built-database) example.

Original JIRA ticket https://issues.couchbase.com/browse/DOC-4068

borrrden:
There is no single way to do this across .NET platforms, so it is a rather complicated question.  I feel like it's not our job to teach people how to get a hold of their certificates.

jamesnocentini:
@borrrden ok understood. I've added a note instead.

Sandychuang8:
Are we closing this issue and not merge the change?

jamesnocentini:
I was getting merge conflicts. Merging now.

PR #112, opened 2019-07-22
DOC-5409 - application events on Android

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5409

### DIFF
--- a/modules/ROOT/pages/java.adoc
+++ b/modules/ROOT/pages/java.adoc
@@ -865,14 +865,13 @@ NOTE: The replication change object also has properties to track the progress (`
 But since the replication occurs in batches and the total count can vary through the course of a replication, those progress indicators are not very useful from the standpoint of an app user.
 Hence, these should not be used for tracking the actual progress of the replication.
 
-==== Replication Status and App Life Cycle
+anchor:repstatus-and-lifecycle[]
 
-Couchbase Lite doesn't react to OS backgrounding or foregrounding events and replication(s) will continue running as long as the remote system does not terminate the connection and the app does not terminate.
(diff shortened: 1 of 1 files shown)


bmeike:
Whoah!  Is this still a thing?  Huge apologies for missing it for so long.
There are no such things as "background" or "foreground" events in Android.  There is also no such thing as an Android app "being in background"
If you can tell me what point you are trying to get across, here I might be able to help you word it....

bmeike (dismissed):
This new section is really neither correct nor useful.  We should try to re-word it.

jamesnocentini:
@bmeike We're trying to explain how one would go about stopping/starting the replicator when the user leaves the app/returns to it, with the current version of Couchbase Lite. How would you recommend that we do it?

bmeike:
Got it.

There really isn’t any equivalent, for Android.  The user is either looking at the app, or she is not.  That is the only programmatically available information.

-blake

bmeike:
It is my opinion that that note only perpetuates the misunderstanding of anyone that thinks it is useful.  Acting on the advice that it gives will result in code that is worse, not better.  I would remove it.

ibsoln:
@rajagp If you're ok with @bmeike comments then I'll remove this change.

ibsoln:
Thanks @bmeike I’ll go with that (https://issues.couchbase.com/browse/DOC-5409) it provides both the warning and a basic avoidance use-case. Hopefully, @rajagp you agree? 
See commit 551beaf on PR#112

rajagp (approved):

bmeike (approved):
Great!

PR #122, opened 2019-08-13
Fix for DOC-5562

bmeike:
update supported api/abi table


modules/ROOT/pages/java.adoc
```diff
@@ -87,6 +87,11 @@ xref:sync-gateway::getting-started.adoc#installation[Installing Sync Gateway ->]
 |Android
 |x86
 |22
+
+|Android
+|x86_64
+|22
+|===
```
jamesnocentini, line 94:
```suggestion
```

PR #123, opened 2019-08-14
Revert "DOC-5578 - restore Android EE internal maven instructions"

jamesnocentini:
This reverts commit 394cdb8031c19dfb45165880d899a21b0bd776cf.

### DIFF
--- a/modules/ROOT/pages/java.adoc
+++ b/modules/ROOT/pages/java.adoc
@@ -31,22 +31,7 @@ dependencies {
 
 ==== Couchbase Lite Enterprise Edition
 
-* In the top-level *build.gradle* file, add the following Couchbase Maven repository in the `allprojects` section.
-+
-[source,groovy]
-----
-allprojects {
-    repositories {
-        google()
-        jcenter()
-        maven {
-            url "https://mobile.maven.couchbase.com/maven2/dev/"
-        }
-    }
-}
-----
-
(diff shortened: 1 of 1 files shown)


jamesnocentini:
@bmeike can this be merged? are we sure it works without it now?

bmeike:
I don't understand how we are ever going to get people to stop using our unsupported, internal server if we keep advertising it.
The short answer to your question is that it will break.  Anyone who is using that URL will break, when we stop supporting it (if that ever happens.  I despair, but that's another story).  In order to break the cycle, it seems to me that we ought to stop advertising it.
Hey!  Just me.
I think @rajagp  has a plan.  I think the best thing to do is to check in with her.

jamesnocentini:
> In order to break the cycle, it seems to me that we ought to stop advertising it.

Sure, as long as it works without it. Otherwise people won’t know how to install EE which is basically like saying “here’s EE, oh and by the way you can’t use it because some instructions are gone”. I’ll merge this PR once we are certain that EE can be installed without the internal maven url.

bmeike (commented):
I've stated my objections.  We should not advertise our flaky as hell internal maven server.

jamesnocentini:
> I've stated my objections. We should not advertise our flaky as hell internal maven server.

@bmeike Noted. Is it currently possible to install EE w/o the internal maven server?

ibsoln (approved):

PR #125, opened 2019-08-20
Doc 5504: Code for DatabaseConfiguration/CouchbaseLite.init API change

bmeike:
Before you flip out, this is two separate commits, so you can cherry-pick appropriately.

### DIFF
--- a/modules/ROOT/examples/java/app/src/main/java/com/couchbase/code_snippets/Examples.java
+++ b/modules/ROOT/examples/java/app/src/main/java/com/couchbase/code_snippets/Examples.java
@@ -29,12 +29,13 @@ import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
-import com.couchbase.lite.AbstractReplicator;
 import com.couchbase.lite.ArrayFunction;
 import com.couchbase.lite.BasicAuthenticator;
 import com.couchbase.lite.Blob;
+import com.couchbase.lite.CouchbaseLite;
 import com.couchbase.lite.CouchbaseLiteException;
 import com.couchbase.lite.DataSource;
 import com.couchbase.lite.Database;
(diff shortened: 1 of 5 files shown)


modules/ROOT/examples/java/app/src/main/java/com/couchbase/code_snippets/Examples.java
```diff
@@ -240,13 +247,14 @@ public void testPreBuiltDatabase() throws IOException {
...
-        DatabaseConfiguration configuration = new DatabaseConfiguration(context);
+        DatabaseConfiguration configuration = new DatabaseConfiguration();
         if (!Database.exists("travel-sample", context.getFilesDir())) {
             ZipUtils.unzip(getAsset("travel-sample.cblite2.zip"), context.getFilesDir());
             File path = new File(context.getFilesDir(), "travel-sample");
             try {
                 Database.copy(path, "travel-sample", configuration);
-            } catch (CouchbaseLiteException e) {
+            }
+            catch (CouchbaseLiteException e) {
```
jamesnocentini, line 257: Is adding a line-break before `catch` a preferred syntax for Android developers? Did you configure Android Studio to show a warning when there isn't a line-break?

I've not seen this syntax before. Anything that adds lines unnecessarily in a code block makes it longer and harder to grasp (personal preference, no science to prove it). So I'm not in favour of this change unless there's a good reason to break the `} catch {` in 2 lines. Being pedantic I know but I believe that stuff really makes a difference considering code snippets get copy pasted often.

bmeike:
The line break is standard Java formatting.  It comes with Android Studio.  Further, I can cite Kevlin Henney.  ... I mean, why the line break after the `try`.  But hey... this code doesn't compile and would crash in an Android app if it did compile.  What's a little formatting?  Follow your instincts!

jamesnocentini:
> The line break is standard Java formatting. It comes with Android Studio.

Ah good to know! Ok let's revisit that line-break change for after 2.6.

Cherry-picked all other changes to 2.5/2.6.

PR #134, opened 2019-10-22
DOC-5710 - finding a database file

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5710

@bmeike @rajagp 2.7 question here, where will CBL Java store the db file? Should we document it differently for Desktop and Server (Tomcat)?

### DIFF
--- a/modules/ROOT/examples/java-platform/src/com/couchbase/code_snippets/Examples.java
+++ b/modules/ROOT/examples/java-platform/src/com/couchbase/code_snippets/Examples.java
@@ -82,4 +82,10 @@ public class Examples {
     }
     // end::getting-started[]
 
+    public void testPrebuiltDatabase() {
+        // tag::prebuilt-database[]
+        // TBA
+        // end::prebuilt-database[]
+    }
+
 }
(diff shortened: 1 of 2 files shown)


bmeike:
This is probably a question for @pasin .  There are two very distinct CBL-Java applications: Desktop and AppServer.  AppServer is more like Android than it is like Desktop.  I suggest that AppServer default to the WebServer's log directory.  The Desktop version might want to use /tmp or something like that...

ibsoln:
> @bmeike @rajagp 2.7 question here, where will CBL Java store the db file? Should we document it differently for Desktop and Server (Tomcat)?
> This is probably a question for @pasin . There are two very distinct CBL-Java applications: Desktop and AppServer. AppServer is more like Android than it is like Desktop. I suggest that AppServer default to the WebServer's log directory. The Desktop version might want to use /tmp or something like that...

@pasin Hi. Do you have a view on the above?

pasin:
I don't have the accurate info about where the default location for the App Server (tomcat) is. I need to find out. You can assign this issue to me to find out.

ibsoln:
@rajagp  @pasin  still needs clarification

I put this in for now:

Databases are created in a directory defined using the API DatabaseConfiguration class as shown in {url-api-references}/com/couchbase/lite/DatabaseConfiguration.html.

Use the Get/Set methods to read/write as needed.

ibsoln:
Withdrawing as the changes here have been subsumed into a larger set of changes to be submitted as a new PR

PR #136, opened 2019-11-11
DOC-5833 - notifications on custom executor on Android

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5833

### DIFF
--- a/modules/ROOT/pages/java.adoc
+++ b/modules/ROOT/pages/java.adoc
@@ -1179,7 +1179,7 @@ xref:index.adoc[Read more]
 *Enhancements*
 
 * Expose the Document Revision ID: the `revisionID` property on a `Document` instance now returns the current revision identifier.
-* The handling of replicator notifications has been enhanced to use a single Executor and 2 x CPU threads.
+* The handling of replicator notifications has been enhanced to use a single executor.
 Apps can specify custom executors in order to override the default behavior (see <<replicator-notifications-on-a-custom-executor>>).
 
 *Android API Support*
--- a/modules/ROOT/pages/java.adoc
+++ b/modules/ROOT/pages/java.adoc
@@ -842,6 +842,154 @@ The level of compression is set on Sync Gateway and can be tuned in the configur
 To optimize for speed, the replication protocol doesn't guarantee that documents will be received in a particular order.
 So we don't recommend to rely on that when using the replication or database change listeners for example.
 
+==== Replicator Notifications on a Custom Executor
+
+Prior to version 2.6, Couchbase Lite spun up multiple executors.
+This policy could result in too many threads being spun up.
+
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/java.adoc
```diff
@@ -842,6 +842,154 @@ The level of compression is set on Sync Gateway and can be tuned in the configur
 To optimize for speed, the replication protocol doesn't guarantee that documents will be received in a particular order.
 So we don't recommend to rely on that when using the replication or database change listeners for example.
 
+==== Replicator Notifications on a Custom Executor
+
+Prior to version 2.6, Couchbase Lite spun up multiple executors.
+This policy could result in too many threads being spun up.
+
+An executor manages a pool of threads and, perhaps, a queue in front of the executor to handle the asynchronous callbacks.
```
bmeike: comma after "executor":  "a queue in front of the executor, to handle...".  It is the pool of threads *and* the queue that handle.

modules/ROOT/pages/java.adoc
```diff
@@ -842,6 +842,154 @@ The level of compression is set on Sync Gateway and can be tuned in the configur
...
+Database.addChangeListener
+Database.addDocumentChangeListener
+Database.addDatabaseChangeListener
+Database.addChangeListener
+----
+
+As of version 2.6, Couchbase Lite spins up a single executor and a number of threads proportional to the number of CPUs `2 x CPUs`.
+
+Note that in some applications with other processes running in parallel, you may see `RejectedExecutionExceptions` error and the Couchbase Lite listener being aborted.
+This happens when there are more than `2 x CPUs` events being processed at the same time.
```
bmeike: This is factually incorrect.
Also, our exact policy is *NOT* something to which we want to commit in writing.
You probably should say something like:

Couchbase sometimes uses its own internal executor to run asynchronous client code.  While this is fine for small tasks, larger tasks -- those that take significant compute time, or that perform I/O -- can block Couchbase processing.  If this happens your application will fail with a `RejectedExecutionException` and it may be necessary to create a separate executor on which to run the large tasks.

bmeike (changes requested):
Need to remove the para that describes our internal threading policy.  We do not want that to be a fixed part of our API.

modules/ROOT/pages/java.adoc
jamesnocentini:
```suggestion
An executor manages a pool of threads and, perhaps, a queue in front of the executor, to handle the asynchronous callbacks.
```

modules/ROOT/pages/java.adoc
jamesnocentini: https://github.com/couchbase/docs-couchbase-lite/pull/136/commits/a5fcdae838dbdb792665e65712761e3d86c40af3 @bmeike Thanks for the re-worked text.

bmeike (approved):
Thanks!

PR #138, opened 2019-11-15
DOC-5618 - Update macOS min supported version

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5618

### DIFF
--- a/modules/ROOT/pages/objc.adoc
+++ b/modules/ROOT/pages/objc.adoc
@@ -133,7 +133,7 @@ include::{partialsdir}/ios-framework-size.adoc[]
 
 |iOS |9.0
 
-|macOS |10.9
+|macOS |10.11
 |===
 
 == API References
@@ -1107,6 +1107,10 @@ The Couchbase Lite API is thread safe except for calls to mutable objects: `Muta
 * Predictive Query
 +
 xref:index.adoc[Read more]
+
+*macOS Support*
+
+Support for macOS 10.9 and 10.10 is being deprecated in this release and will be removed in a future release (see <<supported-versions>>).
 
 *Performance Improvements*
 
--- a/modules/ROOT/pages/objc.adoc
+++ b/modules/ROOT/pages/objc.adoc
@@ -1110,7 +1110,11 @@ xref:index.adoc[Read more]
 
 *macOS Support*
 
-Support for macOS 10.9 and 10.10 is being deprecated in this release and will be removed in a future release (see <<supported-versions>>).
+Support for macOS 10.9 and 10.10 is being deprecated in the 2.6 release and will be removed after two (non-maintenance) releases following the deprecation announcement. 
+--
+
+
+(see <<supported-versions>>).
 
 *Performance Improvements*
 
--- a/modules/ROOT/pages/swift.adoc
+++ b/modules/ROOT/pages/swift.adoc
@@ -1090,7 +1090,7 @@ xref:index.adoc[Read more]
 
 *macOS Support*
 
-Support for macOS 10.9 and 10.10 is being deprecated in this release and will be removed in a future release (see <<supported-versions>>).
+Support for macOS 10.9 and 10.10 is being deprecated in the 2.6 release and will be removed after two (non-maintenance) releases following the deprecation announcement.  (see <<supported-versions>>).
 
 *Performance Improvements*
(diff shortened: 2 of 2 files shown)


modules/ROOT/pages/objc.adoc
```diff
@@ -1108,6 +1108,10 @@ The Couchbase Lite API is thread safe except for calls to mutable objects: `Muta
 +
 xref:index.adoc[Read more]
 
+*macOS Support*
+
+Support for macOS 10.9 and 10.10 is being deprecated in this release and will be removed in a future release (see <<supported-versions>>).
```
rajagp:
```suggestion
Support for macOS 10.9 and 10.10 is being deprecated in the 2.6 release and will be removed after two (non-maintenance) releases following the deprecation announcement. 
--


(see <<supported-versions>>).
```

modules/ROOT/pages/swift.adoc
```diff
@@ -1088,6 +1088,10 @@ The Couchbase Lite API is thread safe except for calls to mutable objects: `Muta
 +
 xref:index.adoc[Read more]
 
+*macOS Support*
+
+Support for macOS 10.9 and 10.10 is being deprecated in this release and will be removed in a future release (see <<supported-versions>>).
```
rajagp:
```suggestion
Support for macOS 10.9 and 10.10 is being deprecated in the 2.6 release and will be removed after two (non-maintenance) releases following the deprecation announcement.  (see <<supported-versions>>).
```

rajagp (changes requested):
Please see comments. Also, assuming that the deprecation notice is for release 2.6.

modules/ROOT/pages/swift.adoc
rajagp: @jamiltz Also, thinking a little more, I think we should replace "in this release"in the suggested line with "version 2.6" . That way this statement would also apply to 2.7 and beyond.

modules/ROOT/pages/objc.adoc
rajagp: @jamiltz  Also, thinking a little more, I think we should replace "in this release"in the suggested line with "version 2.6" . That way this statement would also apply to 2.7 and beyond.

modules/ROOT/pages/objc.adoc
```diff
@@ -1108,6 +1108,14 @@ The Couchbase Lite API is thread safe except for calls to mutable objects: `Muta
 +
 xref:index.adoc[Read more]
 
+*macOS Support*
+
+Support for macOS 10.9 and 10.10 is being deprecated in the 2.6 release and will be removed after two (non-maintenance) releases following the deprecation announcement.
```
eric-schneider, line 1113: @rajagp @ibsoln I find this deprecation notice confusing. It says that the deprecation is being made at 2.6, but this change is being merged into the 2.5 docs. Shouldn't this change be made in the 2.6 release notes?

And to be clear, if this notice is being made at 2.6, then in the following scenario I would expect it to mean the following:
* 2.6 (supported, deprecation announced)
* 2.7 (supported)
* 2.8 (supported)
* 2.9 (UNSUPPORTED)

eric-schneider (changes requested):

modules/ROOT/pages/objc.adoc
ibsoln: I took it to be 'advance' warning ... albeit retrospective. Or maybe it was pushed to wrong release in error. 

@rajagp @eric-schneider  Please advise change needed. Merge to 2.6 or re-word to make it clearer?

rajagp: Oh dear...Its been a while since ticket was opened so I don't even recall! In any case, as of 2.5, the min supported version is 10.11. Which means 10.9 and 10.10 was deprecated in 2.5. Also note that macOS is only supported for testing and dev purposes...always.
2.5 (supported, deprecation announced). Yes, add to 2.5 release notes as well as Eric suggested 
2.6 (supported)
2.7 (supported)
2.8 (UNSUPPORTED)

ibsoln:
Closed and replaced by reworded version PR#172 https://github.com/couchbase/docs-couchbase-lite/pull/172 on 24Jan2020 by @ibsoln

PR #140, opened 2019-11-19
DOC-5994 - supportability android

ibsoln:
https://issues.couchbase.com/browse/DOC-5994

### DIFF
--- a/modules/ROOT/pages/java.adoc
+++ b/modules/ROOT/pages/java.adoc
@@ -74,7 +74,8 @@ xref:sync-gateway::getting-started.adoc#installation[Installing Sync Gateway ->]
 == Supported Versions
 
 The operating systems listed below refer to "Certified" versions of Android.
-We do not test or guarantee the support of uncertified versions such versions of Android built from source.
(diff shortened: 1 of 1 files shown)


RichardSmedley (changes requested):
Maybe change
>> We do not test or guarantee the support of uncertified versions such versions of Android built from source.
to 
>> We do not test or provide support for uncertified versions of Android, such as versions built from source.

bmeike:
I don't think that is correct.  We *will* attempt to support them.  We just don't guarantee that we can.  In particular, when a paying customer puts our code on their custom tablet, we definitely will support it.

RichardSmedley:
> I don't think that is correct. We _will_ attempt to support them. We just don't guarantee that we can. In particular, when a paying customer puts our code on their custom tablet, we definitely will support it.

Ah, in that case I suggest:
_We do not test against, nor guarantee to be able to support, uncertified Android versions such as versions built from source._

bmeike:
Lovely!  Perhaps "... nor guarantee support for ..."

RichardSmedley (approved):
Thanks @ibsoln 
LGTM :+1:

bmeike (approved):

PR #145, opened 2019-12-11
Revert "DOC-5617-x86_64-compatibility"

ibsoln:
Reverts couchbase/docs-couchbase-lite#141


ibsoln:
Backed out as it created a Compare Pull Request message against 2.5 that I was unsure of. Thinking I may have made a mistake and based change on wrong base-branch initially. 

I still don't seem to be able to do review so can you  approve this @RichardSmedley?

PR #154, opened 2020-01-08
DOC-5710-ApplyReviewerChanges-200108-IB

ibsoln:
Apply some corrections spotted by @ibsoln  on 8 Jan 2020 [AM]

### DIFF
--- a/modules/ROOT/examples/java/src/com/couchbase/code_snippets/Examples.java
+++ b/modules/ROOT/examples/java/src/com/couchbase/code_snippets/Examples.java
@@ -23,6 +23,8 @@ private static final String DB_NAME = "gettingstarted";
 /*      Credentials declared this way purely for expediency in this demo - use OAUTH in production code */
 private static final String DB_USER = "sync_gateway";
 private static final String DB_PASS = "password"; // <3>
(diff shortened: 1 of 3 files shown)


modules/ROOT/examples/java/src/com/couchbase/code_snippets/Examples.java
```diff
@@ -23,6 +23,8 @@
 /*      Credentials declared this way purely for expediency in this demo - use OAUTH in production code */
 private static final String DB_USER = "sync_gateway";
 private static final String DB_PASS = "password"; // <3>
+private static final String DB_PATH=new File("").getAbsolutePath()+"/resources";
```
RichardSmedley: Perhaps should be spaces around the =

RichardSmedley (approved):
Other than the two "=" without spaces, and subject to the limits of my mobile Java knowledge, LGTM :+1:

modules/ROOT/examples/java/src/com/couchbase/code_snippets/Examples.java
ibsoln: Indeed

PR #158, opened 2020-01-16
 DOC-6144-Revise-Java-JVM-part2

ibsoln:
Revamped the Getting Started page for Java JVM


ibsoln:
@rajagp @amarantha-k  can this be signed-off? If possible @amarantha-k can you also do a staging build for it?

modules/ROOT/examples/java/src/com/couchbase/code_snippets/Examples.java
```diff
@@ -139,132 +136,531 @@ public static void main (String [] args) throws CouchbaseLiteException, Interrup
...
-                    thisResult.getDouble(Prop_Version)
-                    );
-            outputMessage(resultLine);
-        }
-        outputMessage( String.format("Total rows returned by query = %d", NUMROWS));
+                    thisResult.getDouble(Prop_Version),
+                    thisResult.getString(Prop_Type)));
+        }        outputMessage( String.format("Total rows returned by query = %d", NUMROWS));
+
+        syncTotal = initiateReplication(database, SYNC_GATEWAY_URL, DB_USER, DB_PASS);
```
pasin, line 302: In general, we would recommend users to start the continuous replicator only once, probably when the servlet or app got created. Otherwise, there will be multiple of the same replicator created and step to each other toe when syncing.  It may be fine to provide an example code like this but we probably need to make note about it so that users would know.

modules/ROOT/examples/java/src/com/couchbase/code_snippets/Examples.java
```diff
@@ -139,132 +136,531 @@ public static void main (String [] args) throws CouchbaseLiteException, Interrup
...
         String Prop_Language = "language";
         String Prop_Type = "type";
         String Prop_Version = "version";
+        String searchStringType="SDK";
 
         // Initialise Couchbase
-        outputMessage("== Opening DB and adding record");
 
+//      Initiate CouchbaseLite
         CouchbaseLite.init();
```
pasin, line 242: Normally we would recommend to call init only once when the Web App starts. Calling it multiple time would be fine as well but this means that users will need to remember to do this on every single servlet they have that uses Couchbase Lite. It might be worth to mention about this in the documentation.

modules/ROOT/examples/java/src/com/couchbase/code_snippets/Examples.java
ibsoln: No problem. I'll change this, and include a note to that effect.
For expediency, if I do a non-continuous replication in the app, does that avoid the issue here?

ibsoln:
Revision has been overtaken by other changes to these pages.

PR #169, opened 2020-01-22
Add page alias for java.adoc

amarantha-k:


amarantha-k:
Sorry - the page alias is already on the java-android file, just needs to be uncommented.

PR #177, opened 2020-01-30
DOC-6913 Versions used multiple times in build.gradle should be constants

ibsoln:
https://issues.couchbase.com/browse/DOC-6193


bmeike (approved):
Yah!  That's the ticket!

PR #191, opened 2020-02-20
DOC-6312-Code-Incorrect

ibsoln:
https://issues.couchbase.com/browse/DOC-6312 

Amend code snippets to use correct syntax ConflictResolverProtocol not ConflictResolver


jayahariv (approved):
looks good, we should start building the source after each change, so that this kind of syntactical errors can be caught before submitting.

ibsoln:
> looks good, we should start building the source after each change, so that this kind of syntactical errors can be caught before submitting.

I agree. I'm not sure these snippets changed, which is why they weren't part of the build I did do I think. Going forward I'll do a build regardless.

PR #204, opened 2020-04-02
DOC-6448-Set-java-compatibility-version-in-build-gradle

ibsoln:
https://issues.couchbase.com/browse/DOC-6448


bmeike:
It looks good to me.  I have the following niggles:
1) "Comment out" is very specific.  I think it might be good enough to say "Unnecessary if none of your source code is Kotlin"
2) Consistent code formatting would be nice.  We have
```
kotlinOptions { jvmTarget = '1.8' }
```
and...
```
compileOptions
{
```
and...
```
    maven {
        url "https://mobile.maven.couchbase.com/maven2/dev/"
    }
```
I suggest, instead: 
```
kotlinOptions { jvmTarget = '1.8' }

compileOptions {
...

maven { url "https://mobile.maven.couchbase.com/maven2/dev/" }
```
3) Finally, it would be nice to be consistent about the way sections are displayed.  The point describing what gets included in the `android` section does not show the android section, and is not indented so that it is clear that it goes into the section.  The next two points, #2 & #3 *do* show the including section but do not allude to the fact that the sections will, almost certainly, contain other things, needed by the application program.  Also, note, FWIW, that there is nothing about CBLite that requires either `google()` or `jcenter()`.

These are just niggles.  Adopt them or not as you see fit.

ibsoln:
Content ready to merge. https://couchbase.slack.com/archives/DS1QC61RV/p1585931768001600
Needs an and-1 approval @couchbase/documentation-writers

bmeike (approved):
Looks great!

PR #219, opened 2020-06-12
fix: link error

jayahariv:
* myModel class implementation was missing
* dummy implementation added to avoid the error


jayahariv:
wrong base branch!

PR #223, opened 2020-06-17
DOC-6340-P2P-Using

ibsoln:
Add draft content on using P2P sync in Active and Passive Peers, including code samples


ibsoln:
Admin closure redundant PR no longer relevant

PR #231, opened 2020-09-01
DOC-7292 Add Client Cert authenticator with root certs logic code snippets.

Sandychuang8:
DOC-7292 Add Client Cert authenticator with root certs logic code snippets.


ibsoln (approved):
Thank you

PR #327, opened 2020-10-13
DOC-7513 Remove macOS test only message

ibsoln:
https://issues.couchbase.com/browse/DOC-7513


ibsoln:
Restrict change to Supported-OS page only.

PR #532, opened 2021-09-11
DOC-8272-C4-Android --N1QL Code Snippets

ibsoln:
https://issues.couchbase.com/browse/DOC-8272

### DIFF
--- a/modules/java/examples/code_snippets/Examples.java
+++ b/modules/java/examples/code_snippets/Examples.java
@@ -3142,24 +3142,24 @@ public class TestQueries {
     public void testQueryPagination() throws CouchbaseLiteException {
 
 
-        // tag::query-syntax-pagination[]
-        try {
-          this_Db = new Database("hotels");
-        } catch (CouchbaseLiteException e) {
-          e.printStackTrace();
-        }
+      // tag::query-syntax-pagination[]
+      try {
+        this_Db = new Database("hotels");
+      } catch (CouchbaseLiteException e) {
+        e.printStackTrace();
+      }
 
-        int thisOffset = 0;
-        int thisLimit = 20;
+      int thisOffset = 0;
+      int thisLimit = 20;
 
-        Query listQuery =
-                QueryBuilder
-                        .select(SelectResult.all())
-                        .from(DataSource.database(this_Db))
-                        .limit(Expression.intValue(thisLimit),
-                                  Expression.intValue(thisOffset)); // <.>
+      Query listQuery =
+      QueryBuilder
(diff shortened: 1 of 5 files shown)


modules/java/examples/code_snippets/Examples.java
```diff
@@ -3142,30 +3142,68 @@ public void testQuerySyntaxId() throws CouchbaseLiteException {
...
 
 
-        // tag::query-syntax-pagination[]
-        try {
-          this_Db = new Database("hotels");
-        } catch (CouchbaseLiteException e) {
-          e.printStackTrace();
-        }
+      // tag::query-syntax-pagination[]
+      try {
```
bmeike, line 3146: In Kotlin code, you can probably just leave these try/catch block out. completely.  ... especially ones that don't do any more than print the stack trace and ignore the error

modules/java/examples/code_snippets/Examples.java
```diff
@@ -3142,30 +3142,68 @@ public void testQuerySyntaxId() throws CouchbaseLiteException {
...
+    {
+      // For Documentation -- N1QL Query using parameters
+      // tag::query-syntax-n1ql-params[]
+      //  Declared elsewhere: Database argDB
+
+      Database thisDb = argDB;
+
+      Query thisQuery =
+          thisDb.createQuery(
+              "SELECT META().id AS thisId FROM _ WHERE type = $type"); // <.
```
bmeike, line 3198: Kotlin interprets a dollar sign ("$") inside a string as a variable reference.  This code will fail to compile with a message like "undefined variable 'type'".  The dollar sign needs to be escaped:

"SELECT META().id AS thisId FROM _ WHERE type = \$type")

bmeike (commented):
looks fine with the two noted exceptions

modules/java/examples/code_snippets/Examples.java
ibsoln: Will do

modules/java/examples/code_snippets/Examples.java
ibsoln: This is Java @bmeike so don't need to escape it here do I? But will ensure that it is done in Kotlin module

bmeike: Duh.  This, clearly, *is* Java.  Just ignore the comment.

ibsoln: Dude ... if my Javotlin revolution is winning you over, you ought to ask Jim about my C+-

PR #534, opened 2021-09-11
DOC-8272-C4-Apple --Swift/ObjC N1QL Code Snippets

ibsoln:
https://issues.couchbase.com/browse/DOC-8272

DOC-9047-C4-Code

### DIFF
--- a/modules/objc/examples/code_snippets/SampleCodeTest.m
+++ b/modules/objc/examples/code_snippets/SampleCodeTest.m
@@ -1804,7 +1804,48 @@ CBLQuery* listQuery =
 
 // end::query-syntax-pagination[]
 
+- (void) docsonly_QuerySyntaxN1QL() {
+    /* Documentation Only query-syntax-n1ql
+        Declared elsewhere: CBLDatabase* argDb
+    */
 
+    // tag::query-syntax-n1ql[]
+    NSError *error;
+
+    CBLDatabase *db == argDB;
+
+    NSString *n1qlstr = "SELECT * FROM _ WHERE type = \"hotel\""; // <.>
+
+    CBLQuery* thisQuery = [db createQuery: n1qlstr];
+
+    CBLQueryResultSet* resultset =  [thisQuery execute:&error];
+
+    // end::query-syntax-n1ql[]
+}
+
+- (void) docsonly_QuerySyntaxN1QLParams() {
+    /* Documentation Only query-syntax-n1ql-params
+        Declared elsewhere: CBLDatabase* argDB
+    */
+
+    // tag::query-syntax-n1ql-params[]
+    NSError *error;
+
+    CBLDatabase *db = argDB;
+
+    NSString *n1qlstr = "SELECT * FROM _ WHERE type = $type"; // <.>
+
+    CBLQuery* thisQuery = [db createQuery: n1qlstr];
+
+    CBLQueryParameters* n1qlparams = [[CBLQueryParameters alloc] init];
+    [params setString: @"hotel" forName: @"type"]; // <.>
+
+    thisQuery.parameters = n1qlparams;
+
+    CBLQueryResultSet* resultset =  [thisQuery execute:&error];
+
+    // end::query-syntax-n1ql-params[]
+}
 
 // PEER-to-PEER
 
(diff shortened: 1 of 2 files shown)


modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -1804,7 +1804,48 @@ - (void) dontTestQuerySyntaxJson {
...
 
+- (void) docsonly_QuerySyntaxN1QL() {
+    /* Documentation Only query-syntax-n1ql
+        Declared elsewhere: CBLDatabase* argDb
+    */
 
+    // tag::query-syntax-n1ql[]
+    NSError *error;
+
+    CBLDatabase *db == argDB;
```
jayahariv, line 1813: do you think this two variable declaration are important, and need to be displayed?

I mean, we could keep this above the `tag::query-syntax-n1ql[]`

modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -1804,7 +1804,48 @@ - (void) dontTestQuerySyntaxJson {
...
+
+    NSString *n1qlstr = "SELECT * FROM _ WHERE type = \"hotel\""; // <.>
```
jayahariv, line 1816: not sure, should we promote users to write N1QL query using `_` or with a valid database name?

modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -1804,7 +1804,48 @@ - (void) dontTestQuerySyntaxJson {
...
+
+- (void) docsonly_QuerySyntaxN1QLParams() {
+    /* Documentation Only query-syntax-n1ql-params
+        Declared elsewhere: CBLDatabase* argDB
+    */
+
+    // tag::query-syntax-n1ql-params[]
+    NSError *error;
+
+    CBLDatabase *db = argDB;
```
jayahariv: same, like above.

modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -1804,7 +1804,48 @@ - (void) dontTestQuerySyntaxJson {
...
+
+    NSString *n1qlstr = "SELECT * FROM _ WHERE type = $type"; // <.>
+
+    CBLQuery* thisQuery = [db createQuery: n1qlstr];
+
+    CBLQueryParameters* n1qlparams = [[CBLQueryParameters alloc] init];
+    [params setString: @"hotel" forName: @"type"]; // <.>
+
+    thisQuery.parameters = n1qlparams;
```
jayahariv, line 1842: don't think this is correct! 


But something like the below is a way to set a dynamic variable value to a N1QL query. 
```
NSString* typeVar = @"hotel";
NSString *n1qlstr = [NSString stringWithFormat: @"SELECT * FROM dbName WHERE type = %@", typeVar];
```

jayahariv (changes requested):
please check the comments.

modules/objc/examples/code_snippets/SampleCodeTest.m
ibsoln: DONe

modules/objc/examples/code_snippets/SampleCodeTest.m
ibsoln: As far as I am aware Jay _ is fine to use; may change when collections and scopes are supported

modules/objc/examples/code_snippets/SampleCodeTest.m
ibsoln: Done

modules/objc/examples/code_snippets/SampleCodeTest.m
ibsoln: Hm. Neat trick. Not sure Jay. I thought that we were supposed to use the Query parameters ... maybe both work, but which do we want people using? I'll follow this up and change asap.

jayahariv: @ibsoln  I don't think this code will work, using QueryParams!

PR #536, opened 2021-09-14
DOC-9047-C8-C -- C Replication snippets

ibsoln:
https://issues.couchbase.com/browse/DOC-9047

DOC-9047-C8-C

(cherry picked from commit 21e000795ff0d8f55b7f8d8bf4a9665926e9a83e)

### DIFF
--- a/modules/c/examples/code_snippets/main.c
+++ b/modules/c/examples/code_snippets/main.c
@@ -1,3 +1,4 @@
+// CBL Version 3.0.0 BETA
 #include <cbl/CouchbaseLite.h>
 
 #include <time.h>
@@ -16,14 +17,92 @@ void usleep(unsigned int us) {
 static CBLDatabase* kDatabase;
 static CBLReplicator* kReplicator;
 
+
+//  BEGIN lower-level function declarations
+
+//  DOCS NOTE --
+//  These functions are referred to in subsequent code samples.
+//  Their tags will ensure they are shown alongide the usage examples.
+//  Functions used in more than one place may hve multiple tags.
+
+// tag::p2p-act-rep-func[]
+// tag::p2p-act-rep-add-change-listener[]
+// Purpose -- illustrate a simple change listener
+static void simpleChangeListener(
+                void* context,
+                CBLReplicator* repl,
+                const CBLReplicatorStatus* status) {
+     if(status->error.code != 0) {
+         printf("Error %d / %d\n",
+                status->error.domain,
+                status->error.code);
+     }
+ }
+
+// end::p2p-act-rep-add-change-listener[]
+
+// tag::local-win-conflict-resolver[]
+// Purpose -- illustrate a simple conflict resolver function
+static const CBLDocument* simpleConflictResolver_localWins(
+        void* context, FLString documentID,
+        const CBLDocument* localDocument,
+        const CBLDocument* remoteDocument) {
+        return localDocument;
+    }
+
+// end::local-win-conflict-resolver[]
+// end::p2p-act-rep-func[]
+
+
+
+// tag::replication-push-filter[]
+// tag::replication-pull-filter[]
+// Purpose -- illustrate a simple replication filter function
+static bool simpleReplicationFilter(void* context,
+                                    CBLDocument* argDoc,
+                                    CBLDocumentFlags argFlags)
+{
+    bool result = (argFlags == kCBLDocumentFlagsDeleted);
+    return result;
+}
+
+// end::replication-push-filter[]
+// end::replication-pull-filter[]
+
+
+
+// tag::SimpleReplicationDocumentListener[]
+// Purpose -- Illustrate a simple replication document listener
+static void SimpleReplicationDocumentListener(
+                                              void *context,
+                                              CBLReplicator *replicator,
+                                              bool isPush,
+                                              unsigned numDocuments,
+                                              const CBLReplicatedDocument *documents) {
+
+    if(isPush) {
+        printf("We pushed %d documents",numDocuments);
+    }
+}
+
+// END lower-level function declarations
+
+
+
+
 static void getting_started_change_listener(void* context, CBLReplicator* repl, const CBLReplicatorStatus* status) {
     if(status->error.code != 0) {
         printf("Error %d / %d\n", status->error.domain, status->error.code);
     }
 }
 
+
+// Page=build and run
+// url=https://docs-staging.couchbase.com/couchbase-lite/current/c/gs-build.html
 static void getting_started() {
     // tag::getting-started[]
+    //  Purpose-- provide an overview of available crud  and sync functionality
+    //
     // Get the database (and create it if it doesn't exist)
     CBLError err;
     CBLDatabase* database = CBLDatabase_Open(FLSTR("mydb"), NULL, &err);
@@ -151,6 +230,7 @@ static void getting_started() {
     CBLListener_Remove(token);
     kReplicator = replicator;
 }
+
 
 // tag::local-win-conflict-resolver[]
 static const CBLDocument* local_win_conflict_resolver(void* context, FLString documentID,
@@ -458,7 +538,7 @@ static void database_change_listener() {
     }
     */
 
-    CBLListenerToken* token = CBLDatabase_AddDocumentChangeListener(db, FLSTR("user.john"),
+        CBLListenerToken* token = CBLDatabase_AddDocumentChangeListener(db, FLSTR("user.john"),
         document_listener, NULL);
     // end::document-listener[]
 
@@ -827,7 +907,7 @@ static void order_by() {
     // NOTE: No error handling, for brevity (see getting started)
 
     CBLError err;
-    CBLQuery* query = CBLDatabase_CreateQuery(db, kCBLN1QLLanguage, 
+    CBLQuery* query = CBLDatabase_CreateQuery(db, kCBLN1QLLanguage,
         FLSTR("SELECT meta().id, title FROM _ WHERE type = \"hotel\" ORDER BY title ASC LIMIT 10"),
         NULL, &err);
 
@@ -850,7 +930,7 @@ static void test_explain_statement() {
     // NOTE: No error handling, for brevity (see getting started)
 
     CBLError err;
-    CBLQuery* query = CBLDatabase_CreateQuery(db, kCBLN1QLLanguage, 
+    CBLQuery* query = CBLDatabase_CreateQuery(db, kCBLN1QLLanguage,
         FLSTR("SELECT * FROM _ WHERE type = \"hotel\" GROUP BY country ORDER BY title ASC LIMIT 10"),
         NULL, &err);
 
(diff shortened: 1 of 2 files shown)


ibsoln:
Hi @borrrden can you review the new snippets here. Ignore the p2pmodule for now its content is all duplicated in main for the moment anyhow.

modules/c/examples/code_snippets/main.c
```diff
@@ -16,14 +17,92 @@ void usleep(unsigned int us) {
...
+
+//  BEGIN lower-level function declarations
+
+//  DOCS NOTE --
+//  These functions are referred to in subsequent code samples.
+//  Their tags will ensure they are shown alongide the usage examples.
+//  Functions used in more than one place may hve multiple tags.
+
+// tag::p2p-act-rep-func[]
+// tag::p2p-act-rep-add-change-listener[]
```
borrrden, line 29: Won't this show the tag comment inside of p2p-act-rep-add-change-listener ?

modules/c/examples/code_snippets/main.c
```diff
@@ -1018,7 +1100,44 @@ static void enable_basic_auth() {
...
+    CBLDatabase* database = kDatabase;
+
+    // tag::query-syntax-n1ql-params[]
+    int errorPos;
+
+    CBLError err;
+
+    FLString n1qlstr = FLSTR("SELECT * FROM _ WHERE type = $type");
+
+    FLMutableDict n1qlparams = FLMutableDict_New();
```
borrrden, line 1114: This is never released.  I want to make sure that things like this are released in our doc snippets so that users don't forget to do it.

modules/c/examples/code_snippets/main.c
```diff
@@ -458,7 +538,7 @@ static void database_change_listener() {
     }
     */
 
-    CBLListenerToken* token = CBLDatabase_AddDocumentChangeListener(db, FLSTR("user.john"),
+        CBLListenerToken* token = CBLDatabase_AddDocumentChangeListener(db, FLSTR("user.john"),
```
borrrden: Not a fan of this spacing, it doesn't line up with previous lines.

modules/c/examples/code_snippets/main.c
```diff
@@ -961,6 +1041,8 @@ static void start_replication() {
     kReplicator = replicator;
 }
 
+
+
```
borrrden: Why is so much whitespace being added?

modules/c/examples/code_snippets/main.c
```diff
@@ -1018,7 +1100,44 @@ static void enable_basic_auth() {
...
-    // end::basic-authentication[]
+    // end::basic-authentication-full[]
+}
+
+static void docsonly_N1QL_Params() {
+    CBLDatabase* database = kDatabase;
+
+    // tag::query-syntax-n1ql-params[]
+    int errorPos;
+
```
borrrden, line 1109: Why are there spaces between lines?  Is this to facilitate the overlaid numbers on the rendered snippets?  Otherwise it just wastes space and looks odd.

modules/c/examples/code_snippets/main.c
```diff
@@ -1078,3 +1197,392 @@ Placeholder for Date accessors.
 // tag::fts-index[]
 // placeholder
```
borrrden: I think I've added this in a later commit, FYI

modules/c/examples/code_snippets/main.c
```diff
@@ -1078,3 +1197,392 @@ Placeholder for Date accessors.
...
+    if(docs_example_ShowBasicAuth) {
+        CBLAuthenticator* basicAuth =
+            CBLAuth_CreatePassword(FLSTR("username"),
+                                   FLSTR("passwd"));
+        config.authenticator = basicAuth; // <.>
+    }
+    // end::basic-authentication[]
+
+    // tag::session-authentication[]
+    if(docs_example_ShowSessionAuth) {
```
borrrden, line 1377: What is this `if` condition for?  It won't make sense in the context of a snippet.

modules/c/examples/code_snippets/main.c
```diff
@@ -1078,3 +1197,392 @@ Placeholder for Date accessors.
...
+    CBLError err;
+
+    bool docs_example_resetRequired = argResetRequired;
+    // tag::p2p-act-rep-start-full[]
+
+    CBLReplicator* thisRepl =
+        CBLReplicator_Create(&argConfig, &err); // <.>
+
+    // end::p2p-act-rep-start-full[]
+    if(!docs_example_resetRequired) {
```
borrrden, line 1441: What is all of this about?

modules/c/examples/code_snippets/main.c
```diff
@@ -1078,3 +1197,392 @@ Placeholder for Date accessors.
...
+
+    CBLError err;
+
+    bool docs_example_resetRequired = argResetRequired;
+    // tag::p2p-act-rep-start-full[]
+
+    CBLReplicator* thisRepl =
+        CBLReplicator_Create(&argConfig, &err); // <.>
+
+    // end::p2p-act-rep-start-full[]
```
borrrden, line 1440: Why is there a single line that does not start a replication inside of a tag called "rep-start"?

modules/c/examples/code_snippets/main.c
```diff
@@ -1078,3 +1197,392 @@ Placeholder for Date accessors.
...
+
+    // tag::replication-pendingdocuments[]
+
+    FLDict thisPendingIdList =
+        CBLReplicator_PendingDocumentIDs(thisRepl, &err); // <.>
+
+    if(!FLDict_IsEmpty(thisPendingIdList)) {
+        FLDictIterator item;
+        FLDictIterator_Begin(thisPendingIdList, &item);
+        FLValue itemValue;
```
borrrden, line 1507: This value is never initialized with anything and will surely cause a crash if you tried to run this snippet.  I think you are missing something that is assigning it.

modules/c/examples/code_snippets/main.c
```diff
@@ -1078,3 +1197,392 @@ Placeholder for Date accessors.
...
+// PAGE=Data Sync >> Stop
+// URL=https://docs.couchbase.com/couchbase-lite/current/c/replication.html#lbl-repl-stop
+
+static void docs_act_replication_Stop(
+                                       void* context,
+                                       CBLReplicator* argRepl) {
+    // tag::p2p-act-rep-stop[]
+    // Purpose -- show how to stop a replication
+
+    if(CBLReplicator_Status(argRepl).activity!=kCBLReplicatorStopped) {
```
borrrden, line 1547: Spacing please....

borrrden (changes requested):

modules/c/examples/code_snippets/main.c
ibsoln: No the include now filters out 'tags' by default

modules/c/examples/code_snippets/main.c
ibsoln: Ok. I have done that sporadically. I'll add the releases in.

modules/c/examples/code_snippets/main.c
ibsoln: Habit probably. For a doco whitespace is a good thing. It isolates and highlights and can make things easier to scan read.
As regards code I also tend to group things in 'paragraphs'; grouping related commands. So maybe setting up the args for a function call and the call itself would be one para ... I'll cut it out as much as possible but there will be a difference between 'example' code and how a pro would write production code ... in addition to my Spanglish approach to coding :)

modules/c/examples/code_snippets/main.c
ibsoln: Yes indeed, you did. Thanks

modules/c/examples/code_snippets/main.c
ibsoln: Right. In some examples I want to show two or more ways of doing something, or show off a number of options.

I thought rather than just include them all-in-row -- which would look daft in practice -- I'd include them in some pseudo-conditional that may potentially mimic a viable scenario. I'll revert if you prefer.

modules/c/examples/code_snippets/main.c
ibsoln: The conditionals again? Or the use of &addresses? Or the random use of tags?

Well the include picks up any content in the named tag block, or blocks; so to answer the following query the combined p2p-act-rep-start-full[] snippet does include the start.
What it doesn't include as (confusingly) structured here is the conditional statement syntax (see previous comment about pseudo-conditionals.
I guess I've not decided on approach yet (it's a beta her too).

BTW -- As of the 3.0 our snippets will include a couple click-thru to the github code; so that may change how we present them. Would be happy to get your ideas, Blake's too on how to improve that to better suit engineers.

modules/c/examples/code_snippets/main.c
ibsoln: See previous long-winded reply

modules/c/examples/code_snippets/main.c
ibsoln: Right. Will fix. It ought to be initialized in the 'while; just used 'item' and not `itemvalue` there.

modules/c/examples/code_snippets/main.c
ibsoln: Ok. As regards tags, it can cause a problem if there is no empty line prior to the 'end::' line. So I tend to not have one after the tag:: but include one prior to the end::

I'm sensing you prefer fewer blank lines ... so will try to minimize

ibsoln:
@borrrden I've added in some responses and will update with changes in a subsequent commit.

modules/c/examples/code_snippets/main.c
borrrden: There is a lot of room for style in this, just as you would write an essay for English class.  However, putting so many newlines in like this is the equivalent of writing one sentence paragraphs, which most English teachers would frown on.  I don't think I've ever seen production code written like this, and I was worried that there was some sort of tool sporadically inserting the whitespace when I saw it.  Grouping things together into sort of logical action steps is the normal way I separate things.

modules/c/examples/code_snippets/main.c
borrrden: It would make sense if you were demoing this code to someone to show it function, but if this snippet is getting stuck into docs HTML the first question is going to be "What is docs_example_ShowSessionAuth and why is it here?"

modules/c/examples/code_snippets/main.c
borrrden: Perhaps I should explain that I've left every comment directly below the line I am commenting on unless stated otherwise :).  I'd be wary of clickthrough or presenting this as actual functioning code without testing it first.  It was designed to compile and look good on a page, and that was the extent of the requirement.

modules/c/examples/code_snippets/main.c
borrrden: I understand how the tags work, but what I don't get is have things changed since I last worked on this?  It seems like this would create a one line snippet which tag doesn't seem to match its content.

modules/c/examples/code_snippets/main.c
borrrden: I do prefer fewer blank lines, but what I meant was that there is no space around the `!=` operator.

borrrden:
If you updated then why did this get closed?

PR #610, opened 2021-09-29
Add JSON examples for Kotlin, Android Java and JVM Java

bmeike:
Also moved the Java code down into its proper package.  That Java code is uncompileable, so I don't know how reasonable it is.  We should schedule a task to get it back under compiler control.


ibsoln (approved):
Thanks @bmeike

PR #623, opened 2021-10-08
Update createQuery API

pasin:
Per changes in CBL-2452 that removes the query param tag, update createQuery() API call accordingly.


ibsoln (approved):
Thanks @pasin

PR #715, opened 2022-03-30
update query examples to fix as function

biozal:
The as function requires it to be escaped using ` ` - in the code example there are four times I found the as function without escaping, which means the code won't work for customers that are trying to use this code.  The proper syntax should be:

sudo code follows:
```kotlin
db?.let { database ->
  val query = QueryBuilder
    .select(SelectResult.all())
    .from(DataSource.database(database).`as`("item"))
    .where(Expression.property("type").equalTo(Expression.string("project")))
}
```


ibsoln (approved):
Thank you

PR #717, opened 2022-03-31
update code to include save

biozal:
Updated the code to include save - current inBatch example is missing the save method call which makes the code confusing.


ibsoln (approved):
Thanks

PR #718, opened 2022-03-31
update documentation on indexing

biozal:
The example on ValueIndexConfigurationFactory for creating indexes with Kotlin shows the incorrect way to create an array.  The proper syntax is to use the arrayOf function and pass in the values.


ibsoln (approved):
Thank you

PR #732, opened 2022-07-15
DOC-10214: Ktx Improvements

osfameron:
arrayOf and extension function bits

### DIFF
--- a/modules/android/pages/kotlin.adoc
+++ b/modules/android/pages/kotlin.adoc
@@ -70,6 +70,18 @@ See: https://docs.couchbase.com/mobile/{version-full}/couchbase-lite-android-ktx
 
 {cbl-t} provides a set of {config-factories} and {common-config-factories}, these allow use of named parameters to specify property settings.
 
+This makes it simple to create variant configurations, by simply overriding named parameters:
+
+.Example of overriding configuration
+[source,kotlin]
+----
+val listener8080 = URLEndpointListenerConfigurationFactory.create(
+    networkInterface="en0",
+    port=8080)
+
+val listener8081 = listener8080.create(port=8081)
+----
+
 === Database
 Use
 {database-configuration-factory}
--- a/modules/android/pages/kotlin.adoc
+++ b/modules/android/pages/kotlin.adoc
@@ -176,7 +176,7 @@ In Use::
 ----
 
 db.createIndex("fts_index",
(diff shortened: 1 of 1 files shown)


bmeike (commented):
This is code.... Can we move it out of the text file an into a snippets file?

bmeike (changes requested):
This is code in a documents file.  I would very much prefer it if we never, never did that.  The tagging system seems to be up to dragging in code from wherever.  Let's just use it.

osfameron:
@bmeike yes, let's move it to a snippet.

As that requires scaffolding, and you are reworking the Java samples now in 3.1, does it make sense to:

* merge this PR, but to release/3.1
* you rebase against release/3.1
* you extract the code snippet(s) as part of your current refactor
?

bmeike:
Sigh... I don't think I'm gonna get out of this without more work....  ;-P

Let me suggest an alternative: Can you swap out the code here for tags?  Will the fact that the tags don't exist yet break your build system? 

After you merge this, with the tags, just pass me a ticket to write the code for the tags and I'll be all over it.

bmeike:
Btw... there is existing code that is very like what you want already in the snippets.  Can I suggest that you look at the following tags, in the code to see if they can be used as you need?
- android/examples/kotlin_snippets/app/src/main/kotlin/com/couchbase/codesnippets/ListenerExamples.kt: tag::listener-config-port[], and tag::listener-config-netw-iface[]
- android/examples/kotlin_snippets/app/src/main/kotlin/com/couchbase/codesnippets/QueryExamples.kt: tag::fts-index[]
- android/examples/kotlin_snippets/app/src/main/kotlin/com/couchbase/codesnippets/QueryExamples.kt: tag::query-index[]

If possible, I think it would be great to start consolidating all the code that demonstrates a particular feature, in one place.  I do get it that the key here is the documentation and that we need to do whatever makes the doc work.  Avoiding multiple examples of the same thing, though, I suspect, will make our docs more consistent and more accurate...

bmeike (changes requested):
I have this code.

If you would:
- Remove the code.
- Replace the code with appropriate tags
- Commit
- Open a new ticket to create code for the tags and assign it to me (label "docs", Component Kotlin-Android)
- If you have time. look at the noted places in the snippets and see if any of them, quite similar, can be used instead of creating new snippets.

dinaabennett (approved):
:-)

osfameron:
Closing in favour of #797

PR #733, opened 2022-07-25
DOC-10155: Stopwords

osfameron:
to 3.0
If this is OK, then needs backporting to 2.8


modules/ROOT/pages/_partials/commons/common-fts.adoc
```diff
@@ -39,6 +39,9 @@ FullTextSearch is enabled using the {sqlpp_cbm} match() function.
 
 With the index created, you can construct and run a Full-text search (FTS) query using the indexed properties.
 
+The index will omit a set of common words, to avoid words like "I", "the", "an" from overly influencing your queries.
+See the https://github.com/couchbasedeps/sqlite3-unicodesn/blob/HEAD/stopwords_en.h[full list of these **stopwords**].
```
ThreeJackSparrow: Remove "the", so "See https://github..."

PR #736, opened 2022-08-16
DOC-10320 Add LiteCore Changes to Couchbase-lite 3.0.2 release notes

RayOffiah:
Added missing entries.

### DIFF
--- a/modules/android/pages/_partials/pn-issues-list.adoc
+++ b/modules/android/pages/_partials/pn-issues-list.adoc
@@ -6,11 +6,20 @@
 
 // tag::Fixed-3-0-2[]
 
+
 * https://issues.couchbase.com/browse/CBL-3301[CBL-3301] -- https://issues.couchbase.com/browse/CBL-3301[FullTextIndex.setLanguage() should accept null parameter]
 
 * https://issues.couchbase.com/browse/CBL-3299[CBL-3299] -- https://issues.couchbase.com/browse/CBL-3299[Fix NPE in finalizer]
 
+* https://issues.couchbase.com/browse/CBL-3224[CBL-3224] -- https://issues.couchbase.com/browse/CBL-3224[Call to `c4socket_closed` causes native crash]
+
 * https://issues.couchbase.com/browse/CBL-3136[CBL-3136] -- https://issues.couchbase.com/browse/CBL-3136[Malformed hostname can cause a crash]
+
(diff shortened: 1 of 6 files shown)


modules/android/pages/_partials/pn-issues-list.adoc
```diff
@@ -6,12 +6,21 @@
 
 // tag::Fixed-3-0-2[]
 
+
 * https://issues.couchbase.com/browse/CBL-3301[CBL-3301] -- https://issues.couchbase.com/browse/CBL-3301[FullTextIndex.setLanguage() should accept null parameter]
 
 * https://issues.couchbase.com/browse/CBL-3299[CBL-3299] -- https://issues.couchbase.com/browse/CBL-3299[Fix NPE in finalizer]
 
+* https://issues.couchbase.com/browse/CBL-3224[CBL-3224] -- https://issues.couchbase.com/browse/CBL-3224[Call to `c4socket_closed` causes native crash]
```
osfameron, line 14: I hadn't really registered that you have to write the JIRA issue number *3 times* for these 😱 
(I should get my `jira:` macro merged in at some point maybe...)

osfameron (approved):
Looks good! 👍

modules/android/pages/_partials/pn-issues-list.adoc
RayOffiah: Yes, some of the changes apply to more than one component. 
The LiteCore ones might apply to all of them (look at the list of comments to see which components are affected by a rebuild of the LiteCore module).

osfameron: Ah, separate question, though that's really useful - I think perhaps next time there's a release, I could do it with your guidance? As it feels a little abstract still to me.

But I just meant that for each line, you're having to write e.g. `CBL-3224` 3 times (twice for the ticket number, and once again for the description) which seems like a pain and error prone.
(Just a drive-by 🤔 really, and reminder to self that I wrote that jira: macro and never integrated it)

RayOffiah: Oh, I see! 

Yup, that's quite painful. I just followed the existing pattern, though I'm not really sure why it needs to be two URLs when a single one with some text would have done the trick.

So if you resurrect your JIRA macro that would save a lot of typing.

Yup, I can certainly walk you through it if you need to do one.  

I can explain how it works, but I can't really explain why it was done like this. My best guess that this was part of some automation plan that was going to generate each of the component's issue list files as needed, but it didn't get off the ground.

PR #741, opened 2022-10-19
DOC-10546 Syntax incorrect for Kotlin

RayOffiah:


msecrist-couchbase:
Looks good

osfameron (approved):
👍 

(not sure this is strictly speaking syntax, more like the name of the object/class is wrong, so double checked and yes, actual code examples e.g. https://docs.couchbase.com/couchbase-lite/current/android/querybuilder.html#lbl-return-properties do indeed use `Meta.id` so looks like the change is good in any case.)

PR #742, opened 2022-10-19
DOC-10511 SelectResult seems incorrect in 3.x SDK

RayOffiah:


msecrist-couchbase:
Updates look good. I didn't notice on the changes between lines 39 and 44, there was a line as:
```
The SelectResult returned by `SelectResult.all` is a dictionary object, with the
```
I think the overall problem with this section is some confusion in terminology over what is requested (via SelectResult) and what is returned on the query execution, which is ResultSet. Also, in the above snipet, the code should probably be `SelectResult.all()`

osfameron (approved):

PR #747, opened 2022-11-07
[DOC-10600] Refactor CB mobile release note setup

RayOffiah:


modules/android/pages/releasenotes.adoc
```diff
@@ -8,4 +8,5 @@ ifdef::prerelease[:page-status: {prerelease}]
 
 include::partial$_set_page_context_for_android.adoc[]
 
+include::partial$release-notes/couchbase-mobile-android-release-note.3.0.5.adoc[]
```
osfameron, line 12: so eventually this file will just be a set of includes of the various numbered release notes?

osfameron (approved):
Looks good, much simpler 👍

modules/android/pages/releasenotes.adoc
RayOffiah: That's the plan, yes.
All the release notes (there's one for each module) will just have to create a file and include it at the top, rather than:

Add release notes to the pn_list files
Add tags to the pn_list files
Add variables to the common release notes to make sure that they get included in the right place.

I think this much easier.

osfameron: Ah. Hmm, I wonder if we really even need the complexity of the common release notes and tags 👿 
but yes, much simpler than what we had 👍

RayOffiah: We definitely don't, but common release notes has a bunch of conditional statements relating to each module. It's a bit too much to unpick right now.

PR #753, opened 2022-12-15
Doc 10387 Managing Scopes and Collections

dinaabennett:
Draft
Structure created. Content required.


osfameron (approved):
👍 
Thanks for the authoring pass @dinaabennett!

Note the fixes to Asciidoc (copy/pasting from Google Doc was bit irksome because it "helpfully" expands things like `-` symbols)
and have a look at what I did to make it generic.
(Though can just be curiosity for now! We can go through that process again on a future ticket)

PR #758, opened 2022-12-23
Update all Objective-C code snippets to properly use collections

borrrden:


jayahariv (approved):

modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -265,24 +266,26 @@ - (void) dontTestLoadingPrebuilt {
 
 - (void) dontTestInitializer {
```
osfameron, line 267: I know this hasn't been updated in this commit, but a curiosity: why are all these examples prefixed `dontTest`?

Guessing this has something to do with the test framework used to run this script, but looks a bit odd.
I don't know ObjC, but just to check that someone reading the docs clicks on "View On Github" for context, and sees functions named like this, they'll be familiar with the convention?

osfameron (approved):
Looks great, thanks @borrrden !

PR #767, opened 2023-01-16
CBL-3975: Standardise names, add missing snippet

jayahariv:
```
datatype_usage_createdb		-> database and collection are the names used and are correct. 
datatype_usage_createdoc	-> exists
datatype_usage_mutdict		-> exists
datatype_usage_mutarray		-> exists
datatype_usage_populate		-> exists
datatype_usage_persist 		-> updated to collection.saveDocument()
datatype_usage_closedb		-> exists

date-getter			-> updated the var
datatype_dictionary		-> collection.documentWithJD, updated with doc
datatype_mutable_dictionary	-> collection.save(), 
datatype_array			-> collection.save, doc instead of document
datatype_mutable_array		-> collection.save(), doc instead of document

initializer			-> collection.save(), doc instead of document
update-document			-> collection.documentWithID(), collection.save(), doc instead of document
batch 				-> collection.save(), database instead of self.database

document-listener		-> collection.addDocumentChangeListener(), collection.documentWithID(), doc instead of document. 
document-expiration		-> collection.saveDocExpiration, Datasource.collection(), 

tojson-array			-> array instead of myArray, collection.documentWithID(), Note: No saving since sample only depicts toJSON
tojson-blob			-> collection.documentWithID(), Note: no saving since its only depicts toJSON
tojson-dictionary		-> rename, mydict to dict, 
tojson-document			-> exists, collection.documentWithID(), 


query-access-json		-> exists
query-get-all			-> added the code snippet, it was missing in ObjC.

### DIFF
--- a/modules/objc/examples/code_snippets/SampleCodeTest.m
+++ b/modules/objc/examples/code_snippets/SampleCodeTest.m
@@ -269,11 +269,11 @@
     CBLCollection *collection = [self.database defaultCollection:&error];
 
     // tag::initializer[]
-    CBLMutableDocument *newTask = [[CBLMutableDocument alloc] init];
-    [newTask setString:@"task" forKey:@"task"];
-    [newTask setString:@"todo" forKey:@"owner"];
-    [newTask setString:@"task" forKey:@"createdAt"];
-    [collection saveDocument:newTask error:&error];
+    CBLMutableDocument *doc = [[CBLMutableDocument alloc] init];
+    [doc setString:@"task" forKey:@"task"];
+    [doc setString:@"todo" forKey:@"owner"];
+    [doc setString:@"task" forKey:@"createdAt"];
+    [collection saveDocument:doc error:&error];
     // end::initializer[]
 }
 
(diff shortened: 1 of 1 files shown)


osfameron (approved):
This looks great - again, I'm not a C dev, so I've not verified this works, hopefully you've tested that this runs / or at least compiles.

One question: do you know why the function names are all prefixed "dontTest", like e.g. `dontTestQueryGetAll`

(Mostly a curiosity, but also wonder what the experience will be for a C dev clicking through from the site to view the code on Github. Will they see 'dontTest' and know what that means?)

jayahariv:
* Have verified the code builds successfully. 
* About the prefix, 'dontTest', it was there before I joined, so not 100% sure but what I can guess is, each snippet is added as a unit test, but not allow running them. Prefix with 'test' will make it available for test. IIRC, clicking on snippet will not go to the GH sample code snippets page. If you want to change the function name to a different one than 'dontTest', you could create a ticket, we can update that as well. No preference from my side. @pasin any thoughts from you?

PR #769, opened 2023-01-16
CBL-3974: [swift] variable names are made consistent

jayahariv:
```datatype_usage_createdb		-> 
datatype_usage_createdoc	-> exists
datatype_usage_mutdict		-> exists
datatype_usage_mutarray		-> exists
datatype_usage_populate		-> exists 

datatype_usage_persist		-> collection.save() already updated. 
datatype_usage_closedb		-> exists

date-getter			-> updated doc instead of task
datatype_dictionary		-> collection.document() already updated, doc instead of document. 
datatype_mutable_dictionary	-> mutable_dict -> mutableDict, mutable_doc -> mutableDoc, iOS standards, collection.save() already updated. 
datatype_array			-> mutable_array -> mutableArray, document -> doc
datatype_mutable_array		-> collection.save() already updated, 

initializer			-> collection.save() already updated, newTask -> doc
update-document			-> document->doc, collection.save()

to-dictionary;to-json		-> get document from collection and printed the functions

batch				-> collection save already done. 
document-listener		-> updated with weak collection, collection.addDocumentChangeListener already done. 

document-expiration		-> collection.setDocExpiry, Datasource.collection() already done. 

tojson-array			-> Get doc from collection already done. Save not necessary, since it for demo toJSON. 
tojson-blob			-> same as above 
tojson-dictionary		-> same as above 
tojson-document			-> same as above

query-get-all			-> no change required.
query-access-json		-> exists

Note: mutableDict in case of iOS(objc, swift) code standards, and not 'mutable_dict'

### DIFF
--- a/modules/swift/examples/code_snippets/SampleCodeTest.swift
+++ b/modules/swift/examples/code_snippets/SampleCodeTest.swift
@@ -141,7 +141,7 @@ class SampleCodeTest {
         guard let collection = try self.database.defaultCollection() else {
             fatalError("For sample code snippet, collection should be present!")
         }
-        // tag::ne[]
+        // tag::initializer[]
         let doc = MutableDocument()
             .setString("task", forKey: "type")
             .setString("todo", forKey: "owner")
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/swift/examples/code_snippets/SampleCodeTest.swift [hunk 1]
```diff
@@ -141,12 +141,12 @@ class SampleCodeTest {
         guard let collection = try self.database.defaultCollection() else {
             fatalError("For sample code snippet, collection should be present!")
         }
-        // tag::initializer[]
-        let newTask = MutableDocument()
+        // tag::ne[]
```
osfameron: this looks like a mistake

PR #775, opened 2023-01-18
CBL-4041: [C] Code Samples for FTS and Indexing

velicuvlad:


velicuvlad:
@osfameron

modules/c/examples/code_snippets/main.c
```diff
@@ -1712,26 +1711,56 @@ static void docsonly_N1QL_Params(CBLDatabase* argDb)
...
+    // Set createdAt attribute
+    FLMutableDict_SetString(mutableProperties, FLSTR("createdAt"), FLSliceResult_AsSlice(timeString));
+
+    // Save it to the database
+    if(!CBLCollection_SaveDocument(collection, mutableDoc, &err)) {
+        // Failed to save, do error handling as above
+        return;
+    }
 
-// tag::console-logging[]
```
osfameron, line 1721: What's happened to this one?

modules/c/examples/code_snippets/main.c
```diff
@@ -1712,26 +1711,56 @@ static void docsonly_N1QL_Params(CBLDatabase* argDb)
...
 }
 
+static void date_getter(){
+    CBLDatabase* database = kDatabase;
+    CBLCollection* collection = CBLDatabase_DefaultCollection(kDatabase, NULL);
+    CBLError err;
 
+    // tag::date-getter[]
 
-// tag::console-logging-db[]
```
osfameron, line 1717: Looks like this one has been removed, but that should be OK, as the code snippet is hidden in an `ifndef::is_c`

osfameron (changes requested):

velicuvlad:
in C, if the level of logging is changed, it changes across all domains - i.e cannot be individually set per domain. Snippets of such behaviour are under `tag::logging[]`.

If desired, I can add that tag (`tag::console-logging-db[]`) back but it will remain empty and deleted at some point in future as atm I believe we are looking forward to only keep changing the log level across all domains for all platforms (i.e. current behaviour of C)

osfameron:
Thanks for the clarification @velicuvlad !
I'll merge now, we may need to review some of the text to signal that possibly.

PR #779, opened 2023-01-27
CBL-4025: Working GettingStarted apps for Android, Kotlin, Java Deskt…

bmeike:
This is a substantial change and I have probably broken some things.

Here's the good side:
- There are 4 separate, working applications: Android/Kotlin, Android/Java, JVM/Desktop, JVM/WebService
- the respective gs-build.adoc files now refer to the working code.
- there is only one place in the working code that is an embarrassment.  That bit is commented.
- The getting started code is mostly gone from the snippet_collection files.
- All of the tags in gs-build.adoc point to code that is actually tagged.

... and here's the bad side:
- I may have deleted some tags that are in use (in the snippet_collections files)
- I've probably broken the gs-build.adoc files.  I don't know AsciiDoc heiroglyphics that well (which is surprising since I used it for one of my O'Reilly Books!) and I probably messed something up.  If I had a formatting tool-chain, I'd fix it myself.

I'm more than happy to work with you on getting this in shape.


osfameron (approved):
OK, caveats taken, let's merge it in, and I'll do a full sweep this week of what's happened to all our code examples (e.g. which ones we've broken).

👍

PR #780, opened 2023-01-31
DOC-10793: N1ql To Sql++

osfameron:
NB: using `{sqlpp}` attribute which is common to docs-site playbook rather than the local `{sqlpp_cbm}` attribute previously.


dinaabennett (approved):
Well done!

franrob-projects (approved):
LGTM 👍

PR #785, opened 2023-02-03
CBL-4044: Code examples for CBL replication page

bmeike:
Also prep for pre-commit validation


bmeike:
I've introduced three new tags:

-  tag::execution-inorder[]
-  tag::execution-maxthroughput[]
-  tag::execution-policied[]

... for the three code snippets in the file: `modules/android/pages/_partials/sgw-replication-custom-executor.adoc`

These tags are implemented in both Java and Kotlin.

PR #794, opened 2023-03-08
CBL-4065: More code samples

bmeike:
This is the code for CBL-4065.
It also fixes any use of deprecated code anywhere in any of the snippets.
It all compiles and passes lint tests.
I've tried to make sure all the tags are reasonable but I don't really have good tools for doing that, so there may be mistakes.

Also, the following tags are missing:
- p2p-tlsid-check-bundled
- p2p-tlsid-import-from-bundled
- p2p-tlsid-check-keychain
- p2p-tlsid-store-in-keychain
- p2p-tlsid-return-id-from-keychain

I'm not sure what they are supposed to do.  Is it possible that they describe a feature that is not present in Android/Java.  Let's do a separate ticket for these, if you need them


bmeike:
BTW... Note that tag::logging[] tags a feature that was removed in 3.0.  This is true for all platforms.

osfameron (approved):
👍 thanks Blake for all your hard work on this!
I'm going to merge in, then concentrate on getting the code snippets back into the docs.

PR #800, opened 2023-03-30
Update proguard rules

bmeike:
We have new Proguard rules for v3.1.  I've updated all of the affected projects


osfameron (approved):
Thanks @bmeike !
I know nothing about proguard, but found https://www.guardsquare.com/blog/keep-rules-in-the-age-of-kotlin#
Looks like implementation details on the code, which I'll just merge in. 

Only question: should the docs mention any of this? (As we do a lot of JSON and serialization which interacts with the reflection stuff, will developers expect these rule changes to be *documented*, e.g. in Changelog, or the Data Modeling sections.
Or do they know to head straight for the example rules to copy/paste them, and that's all they need?)

PR #803, opened 2023-04-10
Add missing query-explain code snippets for C

pasin:
Added `query-explain-like`, `query-explain-nopfx`, `query-explain-function`, and `query-explain-nofunction` code snippets for C.


osfameron (approved):
Thanks @pasin

PR #804, opened 2023-04-11
DOC-10991-Review-Comments

dinaabennett:


modules/ROOT/pages/_partials/commons/common-query-n1ql-mobile.adoc
```diff
@@ -765,7 +767,7 @@ SELECT db.*
   WHERE contact.name = "daniel"
 
 SELECT db.contact.address.city
-  FROM db
+  FROM scope.collection
```
pasin, line 770: Remove `db` from SELECT `db.contact.address.city`.

You can also add another example with `collection` like this:

```
SELECT collection.contact.address.city
  FROM scope.collection
```

modules/ROOT/pages/_partials/commons/common-sgw-replication.adoc
```diff
@@ -38,19 +38,15 @@ Each scope can contain one or more collections.
...
 
-You can set up a replication scheme across any of the three data levels:
+You can set up a replication scheme across these data levels:
 
-Database:: All scopes and collections in the database are synced,
-including the `_default` scope and the `_default` collection.
+Database:: The `_default` collection is synced.
 
-Scope:: All the collections in a named scope.
+Collection:: A specific collection is synced.
```
pasin: Could be one or more collections. Maybe `a set of collections`?

modules/ROOT/pages/_partials/commons/common-sgw-replication.adoc
```diff
@@ -515,7 +513,7 @@ Clients wishing to sync previously removed documents must use the resetCheckpoin
 :param-tags: autopurge-override
 include::{root-partials}block_tabbed_code_example.adoc[]
 :param-tags!:
-<.> Here we have opted to turn off the auto purge behavior. By default this is ON (true)
+<.> Here we have opted to turn off the auto purge behavior. By default auto purge is enabled (false)
```
pasin: I think we should remove (false) as the value will depend on the platform. In C, the value will be true and the others will be false.

pasin (changes requested):

modules/ROOT/pages/_partials/commons/common-query-n1ql-mobile.adoc
```diff
@@ -766,7 +766,7 @@ SELECT db.*
   FROM db
   WHERE contact.name = "daniel"
 
-SELECT db.contact.address.city
+SELECT collection.contact.address.city
   FROM scope.collection
   WHERE contact.name = "daniel"
```
pasin: Sorry, I missed thus. For consistency,

```
WHERE collection.contact.name = 'daniel'
```

pasin: Can you add another example without `collection.` as well to avoid the impression that the `collection.` is required?

```
SELECT contact.address.city
FROM scope.collection
WHERE contact.name = "daniel"
```

pasin (approved):
I have a couple comments about the SQL example. The rest looks good to me. Thanks.

PR #810, opened 2023-06-12
DOC-11159: Adds note about non-couchbase support

franrob-projects:
- adds note: Couchbase Lite for Cordova is a community-driven project without official support from Couchbase.
- adds note: Couchbase Lite for React Native is a community-driven project without official support from Couchbase.


osfameron (approved):
Looks fine to me - I see Priya has OK'd this change, though Beth has proposed some updates?
It sounds like there'll be some further discussion and iteration of this in any case, so merge now and update later is fine 👍

PR #822, opened 2023-08-09
Add tag:fleece-data-encoding

bmeike:
also, update tool chains


osfameron (approved):
Looks great, lemme just add a commit to incorporate that tag

PR #829, opened 2023-10-19
DOC-11315: Archive Links

osfameron:
Replace links to archived versions with a link to front page of archive docs.


RichardSmedley (approved):
:+1:

PR #830, opened 2023-10-26
Fix crash bug: premature release of ResultSet

bmeike:
This is a pretty serious bug.
All the way back to 3.0 closing a ResultSet and then accessing a Result that it contains, will cause a native SEGV crash.


osfameron (approved):
thanks Blake!

PR #831, opened 2023-11-06
DOC-11511: CBL-4924 Prereq Install CBL Java

osfameron:
building to https://preview.docs-test.couchbase.com/DOC-11511/couchbase-lite/current/java/gs-prereqs.html


bmeike (commented):
The text is going to work fine: it mostly looks great.
The only thing I see that might need clarification is that this concerns the *runtime* environment, not the *build* environment.  Applications even for Linux, will build with absolutely no problems, even if devs have never heard of the support lib zip.
What they won't do, is run.  Apps that build perfectly will encounter "Cannot load library ...." at runtime, unless the LD_LIBRARY_PATH is set correctly for the process that runs the CBL-Mobile app.

modules/java/pages/gs-prereqs.adoc
```diff
@@ -37,60 +39,63 @@ So, If you have not already done so, see {sgw-prepare--xref}
...
+. Unpack the downloaded file to a location accessible to -- and usable by -- your chosen *build* environment, for example `your_dir/couchbase-lite-java-{version-maintenance-java}`.
 
-. Include the following dependency in your `build.gradle` file, you can remove any {cbl} Maven references:
+. Include the following dependency in your `build.gradle` file:
 +
+[source,gradle,subs="attributes"]
 ----
 Dependencies {
-  implementation fileTree(include: ['*.jar'], dir: <pathToCbl>/lib>
+  implementation fileTree(include: ['*.jar'], dir: your_dir/couchbase-lite-java-{version-maintenance-java})
```
bmeike, line 63: I'm not positive what {version-maintenance-java} is.  It should not be necessary to include a ref to anything in the support lib, in the build.gradle file.

modules/java/pages/gs-prereqs.adoc
```diff
@@ -37,60 +39,63 @@ So, If you have not already done so, see {sgw-prepare--xref}
...
 }
 ----
++
+[#bmkLinuxSharedLibs]
+. Set up the Native Libraries for Linux.
+You will need to add the paths of the directories `<lib-directory>/support/libc++` and `<lib-directory>/support/libicu` to  `LD_LIBRARY_PATH`.
```
bmeike, line 72: I think it is worth mentioning that the thing that actually need to be correct is the Java system property "java.library.path".  That property will usually be set from the shell variable LD_LIBRARY_PATH.

bmeike (approved):
Specific comments

modules/java/pages/gs-prereqs.adoc
ElliotFrancisHunter: It's a decorator defined at the top of this file. Essentially it consists of the major.minor.patch version of java. Do we not need to include the Java version we're targeting in the dependencies?

ElliotFrancisHunter (approved):

modules/java/pages/gs-prereqs.adoc
bmeike: I think I see what is going on, here.  This used to be a reference to an exploded zip.  It should now be a reference to a maven object.  That requires two things:

in setting.gradle (formerly in build.gradle: maybe some people will still have it there) you need to identify our maven server:

    repositories {
        ...
        url "http://mobile.maven.couchbase.com/maven2/dev/"
        ...
    }


Next, replacing the line above in the dependencies (note lower case!) section in build.gradle, you need something like this: 

    dependencies {
        ...
        implementation 'com.couchbase.lite:couchbase-lite-java-ee:3.1.3'
        ...
    }

The "-ee" is the enterprise edition which people should not be using unless they have a license.  The "-3.1.3" is, I think, your "version-maintenance-java"

bmeike: FWIW, to use the CE (community edition) the repositories section just needs `mavenCentral()` (which is probably already there) instead of that `url` line.

osfameron: Yes, the Maven parts are all documented on the gs-install page.

This section is relating only to gs-prereqs, so only for the Linux support libs.

PR #838, opened 2024-01-05
DOC-11757: remove page

ElliotFrancisHunter:
https://issues.couchbase.com/browse/DOC-11757

This is a fix for the issue below:
A user stumbled across this page https://docs.couchbase.com/couchbase-lite/current/c/p2psync-websocket-using-passive.html which really simply shouldn't exist.  This feature is unimplemented in C
(hence all the blank code snippets)

Page had been removed from the nav but not from the repo so users could get there through searching/url hacking.


RichardSmedley (commented):
It might be a good idea to add the page-alias for this page, and the page alias contained within this page, to another page - such as https://github.com/couchbase/docs-couchbase-lite/blob/release/3.1/modules/c/pages/p2psync-websocket.adoc 
...which will stop anyone who has bookmarked this page, waiting for content to appear, from getting a 404.

osfameron:
Agree with @RichardSmedley, did I take you through `:page-aliases:` yet @ElliotFrancisHunter ?

PR #853, opened 2024-03-27
3.2 beta tweak (no prerelease flag)

osfameron:
If applied, this tweak would:

* make "3.2 Beta 1" the default version for mobile
* make it the "current" version
* this means there will be NO "You are viewing docs for a prerelease version" banner
* However, the "Beta 1" orange Page Status badge will show on each docs article page as before.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -1,6 +1,7 @@
 name: couchbase-lite
 version: '3.2'
-prerelease: Beta 1
+display_version: '3.2 Beta 1'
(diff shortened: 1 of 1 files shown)


ElliotFrancisHunter (approved):

osfameron:
Closing, as not all platforms are adopting the 3.2.0 release, so is even more important that we maintain 3.1 as current until then.

PR #854, opened 2024-03-27
Fewd 9785 3.2 beta1 release notes

ElliotFrancisHunter:
Relnotes for Android and iOS 3.2 Beta 1.

Also removed todos for other relnotes and left them as 'none for this release' if people open them.

Added new section specific to the vector search extension in the release notes

### DIFF
--- a/modules/c/partials/release-notes/couchbase-mobile-c-release-note.3.2.0.adoc
+++ b/modules/c/partials/release-notes/couchbase-mobile-c-release-note.3.2.0.adoc
@@ -1,25 +1,42 @@
 [#maint-3-2-0]
-== 3.2.0 -- TODO 2024
+== 3.2.0 -- March 2024
 
 Version 3.2.0 for {param-title} delivers the following features and enhancements:
 
-=== TODO WHAT'S NEW
+NOTE: For an overview of the latest features offered in Couchbase Lite 3.2 Beta 1, see xref:ROOT:cbl-whatsnew.adoc[New in 3.2]
 
-include::ROOT:cbl-whatsnew.adoc[tag=scopes-and-collections]
+== Couchbase Lite Beta 1 Release Notes
 
 === Enhancements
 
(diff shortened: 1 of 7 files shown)


modules/c/partials/release-notes/couchbase-mobile-c-release-note.3.2.0.adoc
```diff
@@ -1,25 +1,42 @@
 [#maint-3-2-0]
-== 3.2.0 -- TODO 2024
+== 3.2.0 Beta 1 -- March 2024
```
osfameron, line 2: I think it sounds like we should NOT have this section, and instead at most a "IMPORTANT: 3.2.0 is not released for this platform" banner instead?

modules/csharp/partials/release-notes/couchbase-mobile-csharp-release-note.3.2.0.adoc
```diff
@@ -1,25 +1,42 @@
 [#maint-3-2-0]
-== 3.2.0 -- TODO 2024
```
osfameron, line 2: Yeah, sorry, think we need to remove these sections, as there is no 3.2 release for these platforms

osfameron (changes requested):
Need to remove from the platfroms that aren't getting it

modules/c/partials/release-notes/couchbase-mobile-c-release-note.3.2.0.adoc
ElliotFrancisHunter: IMPORTANT admonition is a good shout

osfameron (approved):
Better - good to go for now, though I'd remove the 3.2.0 relnote from page entirely, perhaps as 1st iteration after GA

ElliotFrancisHunter:
This is fine for now but requires tidying

PR #855, opened 2024-03-27
DOC-11827: fixed api ref url

ElliotFrancisHunter:
Quick fix until better structural fix tomorrow.

Correctly pointing to API docs.


osfameron (approved):
👍

PR #858, opened 2024-04-11
Doc 12022 spm instructions fix

ElliotFrancisHunter:
Fixed SPM instructions referencing incorrect version (3.1.6 instead of 3.2.0-beta.1)

Caused by incorrect attribute values used.

Expanded scope on the logic that now the attribute was fixed it should be applied to the pages and modules that required it.
Also did some fly-bys

-  Updated minimum API version on android to 23
-  Changed 'Beta 1' text to just 'Beta' as decision was made to change it on docs site rather than beta 1,2,3 as there is no need to have separate pages for the versions.
-  Added attributes for vector search for all sdks to remove hardcoded version for that too as this will be useful for subsequent vector search releases and was needed anyway.
- Changed hardcoded tag attribute in swift, android and objc navs to an include of the relevant module component 
-  Added a previous-version attribute to antora.yml so we don't get a 403 on api reference docs for mobile sdks yet to have a 3.2.0 beta release

### DIFF
--- a/modules/ROOT/pages/_partials/_define_component_attributes.adoc
+++ b/modules/ROOT/pages/_partials/_define_component_attributes.adoc
@@ -11,9 +11,8 @@ ifdef::is_diag[_define_component_attributes.adoc]
 
 // Basic version for Vector Search
 :vs-version: {vs-major}.{vs-minor}
-// Base three digit version including optional tag
-:releasetag: beta.1
-:tag: {empty}
+
+// Base three digit version including optional release tag
 ifdef::releasetag[:tag: -{releasetag}]
 :version-full: {major}.{minor}.{base}{tag}
 :version-full-hyphenated: {major}-{minor}-{base}{tag}
--- a/modules/ROOT/pages/_partials/_define_component_attributes.adoc
+++ b/modules/ROOT/pages/_partials/_define_component_attributes.adoc
@@ -8,7 +8,11 @@ ifdef::is_diag[_define_component_attributes.adoc]
 
 // Basic version
 :version: {major}.{minor}
+
+// Basic version for Vector Search
+:vs-version: {vs-major}.{vs-minor}
 // Base three digit version including optional tag
+:releasetag: beta.1
 :tag: {empty}
 ifdef::releasetag[:tag: -{releasetag}]
 :version-full: {major}.{minor}.{base}{tag}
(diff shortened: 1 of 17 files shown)


modules/ROOT/pages/_partials/_define_component_attributes.adoc
```diff
@@ -8,7 +8,11 @@ ifdef::is_diag[_define_component_attributes.adoc]
 
 // Basic version
 :version: {major}.{minor}
+
+// Basic version for Vector Search
+:vs-version: {vs-major}.{vs-minor}
 // Base three digit version including optional tag
+:releasetag: beta.1
```
osfameron: I'd move this declaration to the antora.yml

e.g. The antora.yml is for things that we *expect to change* at a release, like configuration.

The partial is more like plumbing, to make sure that the things declared in the configuration get formatted in the right way.

modules/ROOT/pages/_partials/_define_component_attributes.adoc
```diff
@@ -18,11 +22,20 @@ ifdef::releasetag[:tag: -{releasetag}]
 // version-maintenance is the full version including maintenance release number.
 // Note that the maintenance release is set at module level in the `_define_module_attributes.adoc` file
 // here at component level it is always same as version-full
-:version-maintenance-android: {major}.{minor}.{maintenance-android}
-:version-maintenance-c: {major}.{minor}.{maintenance-c}
-:version-maintenance-net: {major}.{minor}.{maintenance-net}
-:version-maintenance-java: {major}.{minor}.{maintenance-java}
-:version-maintenance-ios: {major}.{minor}.{maintenance-ios}
+:version-maintenance-android: {major}.{minor}.{maintenance-android}{tag}
```
osfameron, line 24: So, I don't know why these didn't *already* have {tag}...
but as this is a new addition, we'll need to test that this hasn't broken some behavious elsewhere.

modules/ROOT/pages/_partials/_define_component_attributes.adoc
osfameron: not sure that will magically fix it... but it's right to move it to antora.yml, so try that and see?
If not, I'll spend some more time on this tomorrow! Not sure what the problem is from a quick scan.

modules/ROOT/pages/_partials/_define_component_attributes.adoc
ElliotFrancisHunter: It's only used in install, the definition itself and quickstart. Have tested and verified output is correct in all locations.

modules/android/nav-android.adoc
```diff
@@ -1,5 +1,4 @@
-:tag: -beta.1
-ifdef::releasetag[:tag: -{releasetag}]
+include::partial$_set_page_context_for_android.adoc[]
```
osfameron, line 1: You know, I wonder if this line is required?
Given that on the older version of line 48 below, {major} and {minor} etc. *were resolved*...
hmm, I guess those came from the Antora.yml, rather than the page context...

OK, never mind (but I'll try to remember to play with it at some point...)

modules/android/pages/_partials/_define_module_attributes.adoc
```diff
@@ -6,6 +6,12 @@
 // SET full maintennce version number
 :version-maintenance: {version}.{maintenance-android}{tag}
 :version-maintenance-hyphenated: {major}-{minor}-{maintenance-android}{tag}
+
+// VECTOR SEARCH attributes
+:vs-version-maintenance: {vs-version}.{vs-maintenance-android}{tag}
```
osfameron, line 11: It's a bit odd that you use {vs-version} here, but {vs-major}-{vs-minor} directly below.

osfameron (approved):
Looks good!

modules/android/pages/_partials/_define_module_attributes.adoc
ElliotFrancisHunter: It's the hyphens and to mirror the hyphenated version above. I'm not too sure *when* exactly this is used as it's probably an edge case but added it in the off chance I'll need it at some point.

PR #859, opened 2024-04-17
Doc 12113 fix download link

ElliotFrancisHunter:
Fix for the following ticket: https://issues.couchbase.com/browse/DOC-12113

Fixed swift direct download instructions link pointing to CB downloads page which... points to our install docs again.

After some testing across other Vector Search pages, also found the bug in the Obj-C docs and applied the same fix.

Binary is for an iOS download so Swift and Obj-C use the same link.

Fly by fix to Android prereqs docs to fix formatting in which the vector search extension prereqs were unclear due to being added to important admonition instead of separated from it.


osfameron (approved):
👍

PR #860, opened 2024-04-18
CBL-5595 : Add .NET Vector Search Code Snippets

pasin:
* Added .NET Vector Search Code Snippets by translating from Swift Code Snippet.

* Used Color vector / model in the sample code instead of Word.

* Please let me know what needs to be adjusted to be inline with the content in the doc.

### DIFF
--- a/modules/csharp/examples/code_snippets/VectorSearch.cs
+++ b/modules/csharp/examples/code_snippets/VectorSearch.cs
@@ -33,130 +33,96 @@ namespace api_walkthrough
         public string Name { get; set; }
     }
 
-    public sealed class ColorModel : IPredictiveModel
-    {
-        public DictionaryObject Predict(DictionaryObject input)
-        {
-            // Get the input color code 
-            var inputColor = input.GetString("colorInput");
-            if (inputColor == null)
-            {
-                return null;
-            }
-
-            // Get a vector, an array of float numbers, for the input color code.
-            // Normally, you will get the vector from your ML model.
-            float[] vector;
-            try
-            {
-                vector = Color.GetVector(inputColor);
-            }
-            catch (Exception)
-            {
-                return null;
-            }
-
-            // Create an output dictionary by setting the vector result to
-            // the dictionary key named "vector".
-            var retVal = new MutableDictionaryObject();
-            retVal.SetValue("vector", vector);
-            return retVal;
-        }
-
-        public static void Register()
-        {
-            Database.Prediction.RegisterModel("ColorModel", new ColorModel());
-        }
-
-        public static void Unregister()
-        {
-            Database.Prediction.UnregisterModel("ColorModel");
-        }
-    }
-
     public class VectorSearch
     {
-        private readonly Database _database = null;
-        private readonly Collection _collection = null;
-
-        public VectorSearch() {
-            _database = new Database("my-database");
-            _collection = _database.CreateCollection("colors");
-        }
+        public VectorSearch() { }
 
         private static void EnableVectorSearchExtension()
         {
+            // tag::vs - setup - packaging[]
             Extension.Load(new VectorSearchExtension());
+            // end::vs - setup - packaging[]
+        }
+
+        private void CreateDefaultVectorIndexConfig()
+        {
+            // tag::vs - create - default - config[]
+            // Create a vector index configuration for indexing 3 dimensional vectors embedded
+            // in the documents' key named "vector" using 2 centroids.
+            var config = new VectorIndexConfiguration("vector", 3, 2);
+            // end::vs - create - default - config[]
+        }
+
+        private void CreateCustomVectorIndexConfig()
+        {
+            // tag::vs - create - default - config[]
+            // Create a vector index configuration for indexing 3 dimensional vectors embedded
+            // in the documents' key named "vector" using 2 centroids. The config is customized
+            // to use Cosise distance metric, no vector encoding, min training size 100 and
+            // max training size 200.
+            var config = new VectorIndexConfiguration("vector", 3, 2)
+            {
+                DistanceMetric = DistanceMetric.Cosine,
+                Encoding = VectorEncoding.None(),
+                MinTrainingSize = 100,
+                MaxTrainingSize = 200
+            };
+            // end::vs - create - default - config[]
         }
 
         private void CreateVectorIndex()
         {
-            // Create and customize a vector index configuration
-            var config = new VectorIndexConfiguration("vector", 3, 2)
-            {
-                DistanceMetric = DistanceMetric.Euclidean,
-                Encoding = VectorEncoding.None(),
-                MinTrainingSize = 50,
-                MaxTrainingSize = 200
-            };
+            var database = new Database("my-database");
 
-            // Create a vector index from the configuration
-            _collection.CreateIndex("colors_index", config);
-        }
-
-        private void CreateVectorIndexWithEmbedding()
-        {
-            // Create a vector index configuration from a document property named "vector" which
-            // contains the vector embedding.
+            // tag::vs-create-index[]
+            // Create a vector index configuration for indexing 3 dimensional vectors embedded
+            // in the documents' key named "vector" using 2 centroids.
             var config = new VectorIndexConfiguration("vector", 3, 2);
-            _collection.CreateIndex("colors_index", config);
+
+            // Create a vector index named "color_index" using the configuration
+            var collection = database.GetCollection("colors");
+            collection.CreateIndex("colors_index", config);
+            // end::vs-create-index[]
         }
 
-        private IResultSet QueryUsingVectorMatch(string color)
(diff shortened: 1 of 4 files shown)


borrrden (changes requested):
There are no documentation tags on any of the newly added snippets which means they will just exist here for no reason.

borrrden:
Unrelated to this PR but can I delete all other files besides Program.cs and VectorSearch.cs in their directory?  @ElliotFrancisHunter if not, then those files are not getting checked for correctness at all.  @pasin This will probably need to be formatted so that it is readable as standalone snippets (see Program.cs for a reference.  I don't want to have any references to member variables appearing in the snippets, so you will notice that I assign local variables before the doc tag begins).

pasin:
I was informed that the doc team doesn't use tag anymore, and Swift's Vector Search code snippet doesn't use tag. Also to use tag, I would need to know that tag name (provided by the doc team) for each snippet to use.

@ElliotFrancisHunter could you please clarify and provide the direction?

borrrden:
So what...it's just going to plop ALL of that code into the docs?

ElliotFrancisHunter:
> I was informed that the doc team doesn't use tag anymore, and Swift's Vector Search code snippet doesn't use tag. Also to use tag, I would need to know that tag name (provided by the doc team) for each snippet to use.
> 
> @ElliotFrancisHunter could you please clarify and provide the direction?

Hi @pasin and @borrrden!

To clarify, the Swift and Objective-C code now have tags added, I decided to forgo them ahead of beta 1 as we were really pressed for time and I didn't have the bandwidth to learn how to use and include them while still meeting the deadlines.

Going forward we shall still be using tags. Copied across from the ticket:

This contains a list of requirements for code snippets needed in the Couchbase Lite documentation. Show via code snippets how a user can achieve the following:

Configuring a project to use Vector Search.
Any caveats or extras for early access libraries.
Create a default Vector Index Configuration
Create a Vector Index configuration using some optional settings
Creating a Vector Index from embedding
Creating a Vector Index from a Predictive Model
Use vector_match
Use vector_distance
For reference you can look at a similar page on Android: https://docs.couchbase.com/couchbase-lite/3.2/swift/working-with-vector-search.html

Please use the tags:
// tag::vs-create-default-config[]
// tag::vs-create-custom-config[]
// tag::vs-create-index[]
// tag::vs-create-predictive-index[]
// tag::vs-use-vector-match[]
// tag::vs-use-vector-distance[]
// tag::vs-setup-packaging[]

These tags have been selected to mirror existing snippet tags for the other sdks.

osfameron:
@borrrden @ElliotFrancisHunter 

> Unrelated to this PR but can I delete all other files besides Program.cs and VectorSearch.cs in their directory? @ElliotFrancisHunter if not, then those files are not getting checked for correctness at all. @pasin This will probably need to be formatted so that it is readable as standalone snippets 

We (ideally) don't want "standalone snippets", so code that *can be run* would be better.
(And code that *is run*, automatically is best of all.)
But in Docs we're not .NET experts, so defer to you on how any checking is done.

If you change e.g. the paths of code files that have tags, please let us know, so we can update docs.

(One thought: I used .csx files, run with dotnet-script for the .NET Server SDK examples.
I liked it, as it was standalone, and let me run/test examples for arbitrary code snippet files,
even if other ones had bitrotted.
But I don't think the SDK devs liked it, and may be removing it, which is their decision as the .NET experts so 🤷)

RichardSmedley:
> (One thought: I used .csx files, run with dotnet-script for the .NET Server SDK examples. I liked it, as it was standalone, and let me run/test examples for arbitrary code snippet files, even if other ones had bitrotted. But I don't think the SDK devs liked it, and may be removing it, which is their decision as the .NET experts so 🤷)

It was removed as it was not widely used within the industry - nor by the devs here - so the maintainability gain became a maintainability loss (it also consequently broke with some updates to the code samples). 
iow appropriateness of csx entirely depends on the workflow of those who have to maintain it (it just didn't work out for the Server SDKs).

pasin:
@ElliotFrancisHunter @borrrden 
I tagged the code snippet and sync the code with Android's snippet but using color vectors instead. I added a new tag named "tag::vs-predictive-model[]" for the predictive model class snippet.

borrrden:
>We (ideally) don't want "standalone snippets", so code that can be run would be better.
(And code that is run, automatically is best of all.)
But in Docs we're not .NET experts, so defer to you on how any checking is done.

That's my point here.  Someone arbitrarily added new files into the directory that are not getting checked for correctness.  Before this PR the ONLY file that was checked was Program.cs.  Now with this PR VectorSearch.cs is also getting checked.  

https://github.com/couchbase/docs-couchbase-lite/tree/release/3.2/modules/csharp/examples/code_snippets

There are other files in here like JsonApiMethods.cs, QueryViewModel.cs, and p2psync-websocket.cs which are not getting checked at all.  I went through a lot of effort to get Program.cs in a state where it can be compiled, but it can also show up as nice looking snippets in the actual documentation.  They are not just snippets pasted into a file, they are thought out snippets in a file that is actually able to be compiled and I won't be very motivated to help again if it gets into the same uncontrolled state as it used to be where everyone just dumped code into a file wherever they felt like (as you might imagine this led to rot of the snippets exceptionally quickly and documentation errors all over the place).  So my question is can the three files above be deleted?  Or are they being used?  If they are being used, they are NOT being checked for correctness at all.

borrrden (approved):

ElliotFrancisHunter:
> > We (ideally) don't want "standalone snippets", so code that can be run would be better.
> > (And code that is run, automatically is best of all.)
> > But in Docs we're not .NET experts, so defer to you on how any checking is done.
> 
> That's my point here. Someone arbitrarily added new files into the directory that are not getting checked for correctness. Before this PR the ONLY file that was checked was Program.cs. Now with this PR VectorSearch.cs is also getting checked.
> 
> https://github.com/couchbase/docs-couchbase-lite/tree/release/3.2/modules/csharp/examples/code_snippets
> 
> There are other files in here like JsonApiMethods.cs, QueryViewModel.cs, and p2psync-websocket.cs which are not getting checked at all. I went through a lot of effort to get Program.cs in a state where it can be compiled, but it can also show up as nice looking snippets in the actual documentation. They are not just snippets pasted into a file, they are thought out snippets in a file that is actually able to be compiled and I won't be very motivated to help again if it gets into the same uncontrolled state as it used to be where everyone just dumped code into a file wherever they felt like (as you might imagine this led to rot of the snippets exceptionally quickly and documentation errors all over the place). So my question is can the three files above be deleted? Or are they being used? If they are being used, they are NOT being checked for correctness at all.

Hakim and I will look into the extra files on Monday when he's back from PTO, tidying up the snippets and making sure they can be correctly tested is a priority of mine in the long-term uplifting and wholesale tidy up of the mobile docs codebase outside of ongoing docs effort.  

For now, merging this and creating epic to ensure this task is addressed when there is time to do it properly: https://issues.couchbase.com/browse/DOC-12120

modules/csharp/examples/code_snippets/VectorSearch.cs
```diff
@@ -0,0 +1,203 @@
...
+        private static void EnableVectorSearchExtension()
+        {
+            // tag::vs - setup - packaging[]
+            Extension.Load(new VectorSearchExtension());
+            // end::vs - setup - packaging[]
+        }
+
+        private void CreateDefaultVectorIndexConfig()
+        {
+            // tag::vs - create - default - config[]
```
osfameron, line 49: rewrite to:
`// tag::vs-create-default-config[]` e.g. without spaces

modules/csharp/examples/code_snippets/VectorSearch.cs
```diff
@@ -0,0 +1,203 @@
...
+            // Create a vector index configuration for indexing 3 dimensional vectors embedded
+            // in the documents' key named "vector" using 2 centroids.
+            var config = new VectorIndexConfiguration("vector", 3, 2);
+            // end::vs - create - default - config[]
+        }
+
+        private void CreateCustomVectorIndexConfig()
+        {
+            // tag::vs - create - default - config[]
```
osfameron, line 58: presumably this should be `tag::vs-create-custom-config[]` e.g. default -> custom

modules/csharp/examples/code_snippets/VectorSearch.cs
```diff
@@ -0,0 +1,203 @@
...
+                }
+
+                // Create an output dictionary by setting the vector result to
+                // the dictionary key named "vector".
+                var retVal = new MutableDictionaryObject();
+                retVal.SetValue("vector", vector);
+                return retVal;
+            }
+        }
+        // tag::end-predictive-model[]
```
osfameron, line 119: this should be `end::vs-predictive-model[]`
@ElliotFrancisHunter you'll need to go through all the tags/ends and make sure they match

PR #863, opened 2024-05-07
Doc 12181 3.1.7 relnotes

ElliotFrancisHunter:
Release notes for 3.1.7 maintenance release.
Also includes a quick fix to a label in c release notes pointing to the wrong version.
Antora.yml updated so version attributes now say 3.1.7 instead of 3.1.6 for 3.1 docs.

### DIFF
--- a/modules/c/partials/release-notes/couchbase-mobile-c-release-note.3.1.6.adoc
+++ b/modules/c/partials/release-notes/couchbase-mobile-c-release-note.3.1.6.adoc
@@ -1,4 +1,4 @@
-[#maint-3-1-3]
+[#maint-3-1-6]
 == 3.1.6 -- March 2024
 
 Version 3.1.6 for {param-title} delivers the following features and enhancements:
--- a/modules/csharp/pages/releasenotes.adoc
+++ b/modules/csharp/pages/releasenotes.adoc
@@ -9,8 +9,12 @@ ifdef::prerelease[:page-status: {prerelease}]
 include::partial$_set_page_context_for_csharp.adoc[]
 
 [#maint-latest]
+include::partial$release-notes/couchbase-mobile-csharp-release-note.3.1.7.adoc[]
+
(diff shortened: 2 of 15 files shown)


modules/csharp/pages/releasenotes.adoc
```diff
@@ -9,8 +9,12 @@ ifdef::prerelease[:page-status: {prerelease}]
 include::partial$_set_page_context_for_csharp.adoc[]
 
 [#maint-latest]
+include::partial$release-notes/couchbase-mobile-csharp-release-note.3.1.7.adoc[]
+
```
osfameron, line 13: Adding all this spacing is fine as long as it makes sense to you!
But makes me realise I didn't explain the original layout (e.g. that made sense to me) which was:

```
[#maint-latest]
include::...(latest include)

include::...(earlier ones, all bunched together)
include::...
include::...
```

e.g. just a single line that separates the current from the rest.

(Not at all a problem if you like the extra spaces though, just surprised me as I wasn't expecting it, and then I had to remember what I'd done before 🤷)

modules/c/partials/release-notes/couchbase-mobile-c-release-note.3.1.6.adoc
```diff
@@ -1,4 +1,4 @@
-[#maint-3-1-3]
+[#maint-3-1-6]
```
osfameron, line 1: good catch

osfameron (approved):
Looks good to me 👍 (I've not built it to test links, assume you've built preview/local and all looks OK?

modules/csharp/pages/releasenotes.adoc
ElliotFrancisHunter: Yeah I remembered the most recent needs to be tied to the label, just scans a bit easier for me with the spacing and how I've been drilled in the past

PR #869, opened 2024-05-13
CBL-5561: Create Jenkins validation for CBL-iOS DOC snippets

velicuvlad:


osfameron (approved):
This is very much your domain as Engineering, so if you need detailed review, please defer to Pasin and your colleagues!

But I'm happy to wave through assuming you've done diligence 👍

osfameron:
Hmm, on second thoughts, yes, I'm still happy to wave it through, but it's not a bad idea for me to have a better understanding of it... lemme spend another 15 mins looking through it now... 

also happy to discuss if you'd like to take me through any of it for a sense check.

velicuvlad:
this PR is only for iOS. For .NET and C there is already validation in-place.

CBL-iOS snippets do live inside a project app. In order to guarantee that, at least, there are no syntax errors within the snippets we want to build the application. Thus, this PR will add an extra stage in Jenkins (alongside existing .NET and C) so that we always make sure, if any modifications are made to the snippets, that those are correct.
As you already know, iOS doesn't have just a list of snippets, there is some additional setup such as defining the db and collection globally etc. that is not included in the official, published, documentation (encapsulated in tags)

TL;DR: We want to build the application "code_snippets" which is the CBL-iOS application containing all snippets example in order to make sure that, at least, there are no syntax errors.

PR #881, opened 2024-07-16
Doc 12354 fix beta versioning

ElliotFrancisHunter:
This is a fix for the following ticket: https://issues.couchbase.com/browse/DOC-12354

Description: All platforms other than C return a 403 error when trying to access -beta.3 API references pages. They all seem to work for -beta.2. Not sure if this is intended behaviour to have different versions.

This has also broken some links that directly point to API reference pages based on version.

Example: https://docs.couchbase.com/couchbase-lite/current/csharp/database.html#close-database

The issue also resulted in download binaries being incorrect and multiple links being broken across all platforms. 
Additional fly-by fixes:
-- Copy editing
-- Changing file extension of c examples to .cpp to fix c platform examples not being included.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -18,7 +18,7 @@ asciidoc:
     page-status: Beta
     previous-release: 3.1.6
     release: '3.2'
-    releasetag: beta.3
+    releasetag: beta.2
     major: 3
     minor: 2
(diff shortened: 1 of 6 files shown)


osfameron (approved):
Looks good 👍

PR #880, opened 2024-07-16
CBL-5976: LazyIndex .NET snippets

borrrden:

PR #882, opened 2024-07-17
CBL-5969: Add NumProbes to custom vector index config

borrrden:


borrrden:
The validation here is going to depend on the fixes that came with #880

ElliotFrancisHunter:
@borrrden I've merged in #880 - for some reason the validate build seems to still be failing for the remaining c# builds

borrrden:
rebased

ElliotFrancisHunter (approved):

PR #883, opened 2024-07-18
CBL-5968 and CBL-5973 : numProbes and lazy index snippet for C

pasin:
* Added setting numProbes snippet to vs-create-custom-config tag.

* Added lazy index snippet. 

* Updated number of centroids to 100 in all snippets for consistency.

* Changed to use the latest API and SQL syntax.


ElliotFrancisHunter (approved):
This is great! Thank you 👍🏼

PR #887, opened 2024-07-19
Vector Search Code Snippet Update for Objective-C

pasin:
* CBL-5971, CBL-5975, CBL-6020, CBL-6036, CBL-6041.

* New Tags added: vs-setup-packaging, vs-numprobes-config, vs-apvd-prediction, vs-apvd-order-by, tag::vs-apvd-where, vs-hybrid-order-by, vs-hybrid-where,  vs-hybrid-prediction, vs-hybrid-ftmatch, vs-lazy-index-config, vs-create-lazy-index-embedding.

* vs-hybrid-vmatch is not applicable because we don’t have vector_match() anymore.

* Dev feels like the query examples (various of regular and hybrid vector search query) do not need code snippt. They can be listed as an example of the vector search queries using APPROX_VECTOR_DISTANCE. Even though I have implemented in this PR but could you please review them when you have the content written?

### DIFF
--- a/modules/objc/examples/code_snippets/VectorSearch.m
+++ b/modules/objc/examples/code_snippets/VectorSearch.m
@@ -7,34 +7,37 @@
 
 #import <Foundation/Foundation.h>
 #import <CouchbaseLite/CouchbaseLite.h>
-#import <NaturalLanguage/NLEmbedding.h>
 
-// Placeholder for the vector not found error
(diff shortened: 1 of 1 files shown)


ElliotFrancisHunter (approved):
Will review and refine once docs copy has been drafted. Thank you! 👍🏼

PR #888, opened 2024-07-19
Vector Search Code Snippet Update for Swift

pasin:
* CBL-5970, CBL-5974, CBL-6021, CBL-6035, CBL-6040

* New Tags added: vs-setup-packaging, vs-numprobes-config, vs-apvd-prediction, vs-apvd-order-by, tag::vs-apvd-where, vs-hybrid-order-by, vs-hybrid-where,  vs-hybrid-prediction, vs-hybrid-ftmatch, vs-lazy-index-config, vs-create-lazy-index-embedding.

* vs-hybrid-vmatch is not applicable because we don’t have vector_match() anymore.

* Dev feels like the query examples (various of regular and hybrid vector search query) do not need code snippt. They can be listed as an example of the vector search queries using APPROX_VECTOR_DISTANCE. Even though I have implemented in this PR but could you please review them when you have the content written?

### DIFF
--- a/modules/swift/examples/code_snippets/VectorSearch.swift
+++ b/modules/swift/examples/code_snippets/VectorSearch.swift
@@ -12,7 +12,7 @@ var database: Database!
 var collection: Collection!
 
 enum AppError: Error {
-    case indexdNotFound
+    case indexNotFound
     case vectorNotFound
 }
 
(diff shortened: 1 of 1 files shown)


modules/swift/examples/code_snippets/VectorSearch.swift
```diff
@@ -7,16 +7,29 @@
...
 import CouchbaseLiteSwift
-import NaturalLanguage
 
 var database: Database!
 var collection: Collection!
-var mlmodel: NLEmbedding!
 
-enum VectorSearchError: Error {
+enum AppError: Error {
+    case indexdNotFound
```
ElliotFrancisHunter: Noticed a typo here. Should this read indexNotFound?

ElliotFrancisHunter (dismissed):
Happy if that enum case isn't a typo but thought it would be handy for you to double check before I approve and merge. Looks great otherwise :)

velicuvlad (approved):

PR #891, opened 2024-07-25
DOC-12385: Copy across missing snippets

ElliotFrancisHunter:
Move snippets from p2psync-websocket.cs to Program.cs

### DIFF
--- a/modules/csharp/examples/code_snippets/Program.cs
+++ b/modules/csharp/examples/code_snippets/Program.cs
@@ -2450,118 +2450,119 @@ namespace api_walkthrough
         }
     }
     // end::merge-conflict-resolver[]
+
+
+    // tag::p2p-act-rep-func[]
+    // . . . preceding code. for example . . .
+        private static ListenerToken _thisListenerToken;
+        var Database thisDB;
+        // . . . other code . . .
+
+    // tag::p2p-act-rep-config-type[]
+    // Set replicator type
+    thisConfig.ReplicatorType = ReplicatorType.PushAndPull;
+
+    // end::p2p-act-rep-config-type[]
+
+    // tag::autopurge-override[]
+    // Set autopurge option
+    // here we override its default
+    thisConfig.EnableAutoPurge = false; // <.>
+
+    // end::autopurge-override[]
+
+    // tag::p2p-act-rep-config-cont[]
+    // Configure Sync Mode
+    thisConfig.Continuous = true; // default value
+
+    // end::p2p-act-rep-config-cont[]
+
+    // tag::p2p-act-rep-config-self-cert[]
+    // Configure Server Security -- only accept self-signed certs
+    thisConfig.AcceptOnlySelfSignedServerCertificate = true; // <.>
+
+    // end::p2p-act-rep-config-self-cert[]
+    // Configure Client Security // <.>
+
+    // tag::p2p-act-rep-auth[]
+    // Configure basic auth using user credentials
+        thisConfig.Authenticator = new BasicAuthenticator("Our Username", "Our Password");
+
+    // end::p2p-act-rep-auth[]
+
+    // tag::p2p-act-rep-start-full[]
+    // Initialize and start a replicator
+    // Initialize replicator with configuration data
+    var thisReplicator = new Replicator(thisConfig); // <.>
+
+    // tag::p2p-act-rep-add-change-listener[]
+    // tag::p2p-act-rep-add-change-listener-label[]
+    //Optionally add a change listener // <.>
+    // end::p2p-act-rep-add-change-listener-label[]
+    _thisListenerToken =
+        thisReplicator.AddChangeListener((sender, args) =>
+            {
+            if (args.Status.Activity == ReplicatorActivityLevel.Stopped) {
+                Console.WriteLine("Replication stopped");
+            }
+            });
+
+    // end::p2p-act-rep-add-change-listener[]
+
+    // tag::p2p-act-rep-start[]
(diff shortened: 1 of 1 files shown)


borrrden:
This has broken compilation and cannot be merged

borrrden (changes requested):
This needs to be written in a way that matches the existing code in Program.cs.  It _cannot_ just be plopped in.  This is what the GH checks are meant to prevent.  These snippets should be able to be compiled.

osfameron:
Thanks @borrrden - who of the devs can we ask to look at that?
Can you let us know which ticket type to create and link to the DOC ticket above?

borrrden:
Ideally it would be the person who wrote them that puts them in correctly.  Otherwise it's going to be me.  Are all of these snippets needed?  Some of them are just one line which seems a bit excessive to create an entire self contained snippet for.

borrrden (commented):
Would like some clarification here on whether these are all important.  There are a lot of tags in here including "how to call start" which seems pretty obvious, does it really need a snippet?

Anyway perhaps a description of what each one wants to do is in order because some of them are suspicious.  I'll put in comments inline.

modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2455,80 +2452,105 @@ public Document Resolve(Conflict conflict)
     // end::merge-conflict-resolver[]
  }
 
-#warning p2p-act-rep-func used, but contains nothing
 // tag::p2p-act-rep-func[]
+// . . . preceding code. for example . . .
+    private static ListenerToken _thisListenerToken;
+    var Database thisDB;
+    // . . . other code . . .
```
borrrden, line 2459: What is this meant to represent?

modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2455,80 +2452,105 @@ public Document Resolve(Conflict conflict)
...
 // Set autopurge option
 // here we override its default
+thisConfig.EnableAutoPurge = false; // <.>
 
 // end::autopurge-override[]
 
-#warning p2p-act-rep-config-cont used, but contains nothing
 // tag::p2p-act-rep-config-cont[]
 // Configure Sync Mode
+thisConfig.Continuous = true; // default value
```
borrrden, line 2476: This is not the default for this property

modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2455,80 +2452,105 @@ public Document Resolve(Conflict conflict)
...
+              Console.WriteLine("Replication stopped");
+          }
+        });
 
 // end::p2p-act-rep-add-change-listener[]
 
-#warning p2p-act-rep-start used, but contains nothing
 // tag::p2p-act-rep-start[]
 // Start replicator
+ thisReplicator.Start(); // <.>
```
borrrden, line 2514: This is the same as starting any other replicator...does it really need a separate snippet?

modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2455,80 +2452,105 @@ public Document Resolve(Conflict conflict)
...
-#warning p2p-act-rep-config-cacert used, but contains nothing
 // tag::p2p-act-rep-config-cacert[]
 // Configure Server Security -- only accept CA certs
+thisConfig.AcceptOnlySelfSignedServerCertificate = false; // <.>
+
 // end::p2p-act-rep-config-cacert[]
 
-#warning p2p-act-rep-config-cacert-pinned used, but contains nothing
 // tag::p2p-act-rep-config-cacert-pinned[]
 // Only CA Certs accepted
```
borrrden, line 2527: Misleading comment here.  Only certificates issued by this CA are accepted, not any CA cert.

modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2455,80 +2452,105 @@ public Document Resolve(Conflict conflict)
...
+
 // end::p2p-act-rep-config-cacert-pinned[]
 
-#warning p2p-act-rep-status used, but contains nothing
 // tag::p2p-act-rep-status[]
+_thisReplicator.Stop();
+    while (_thisReplicator.Status.Activity != ReplicatorActivityLevel.Stopped) {
+        // Database cannot close until replicators are stopped
+        Console.WriteLine($"Waiting for replicator to stop (currently {_thisReplicator.Status.Activity})...");
+        Thread.Sleep(200);
```
borrrden, line 2544: This is quite domain specific and unlikely to make a helpful snippet

modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2455,80 +2452,105 @@ public Document Resolve(Conflict conflict)
...
+    }
+_thisDatabase.Close();
```
borrrden, line 2546: Why does this need to be here?

modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2455,80 +2452,105 @@ public Document Resolve(Conflict conflict)
...
+
 // end::p2p-act-rep-status[]
 
-#warning p2p-act-rep-stop used, but contains nothing
 // tag::p2p-act-rep-stop[]
 // Stop replication.
+thisReplicator.Stop(); // <.>
```
borrrden, line 2552: Same as any other replication...

ElliotFrancisHunter:
Closed as underlying issue has been fixed. Will still need to move these snippets over at some point.

PR #894, opened 2024-08-14
AV-84097: Fix copy to reflect updated snippet

ElliotFrancisHunter:
Fix for the ticket: https://couchbasecloud.atlassian.net/browse/AV-84097

Snippets have been updated but the copy for Android and Java was not.

### DIFF
--- a/modules/android/pages/working-with-vector-search.adoc
+++ b/modules/android/pages/working-with-vector-search.adoc
@@ -36,10 +36,10 @@ include::java:example$snippets/common/main/java/com/couchbase/codesnippets/Vecto
 
 First, initialize the `config` variable with the `VectorIndexConfiguration()` method with the following parameters:
 
-* The width or `dimensions` of the vector index is set to `300`.
+* The width or `dimensions` of the vector index is set to `3`.
 
-* The amount of `centroids` is set to `20`.
-This means that there will be 20 buckets with a single centroid each that gathers together similar vectors.
+* The amount of `centroids` is set to `100`.
+This means that there will be 100 buckets with a single centroid each that gathers together similar vectors.
 
 There are also some default values that we did not set with the `VectorIndexConfiguration()` method, these are:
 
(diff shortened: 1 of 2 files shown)


modules/android/pages/working-with-vector-search.adoc
```diff
@@ -36,10 +36,10 @@ include::java:example$snippets/common/main/java/com/couchbase/codesnippets/Vecto
 
 First, initialize the `config` variable with the `VectorIndexConfiguration()` method with the following parameters:
 
-* The width or `dimensions` of the vector index is set to `300`.
+* The width or `dimensions` of the vector index is set to `3`.
 
-* The amount of `centroids` is set to `20`.
-This means that there will be 20 buckets with a single centroid each that gathers together similar vectors.
+* The amount of `centroids` is set to `100`.
+This means that there will be 100 buckets with a single centroid each that gathers together similar vectors.
```
RichardSmedley, line 42: Is there somewhere else in the mobile docs that defines `bucket` in opposition to regular Couchbase Server buckets?

ElliotFrancisHunter: There is no Couchbase Lite definition of bucket on the mobile side no, was a term decided by PM

osfameron: "bucket" is a fairly common name in algorithms... but that's a good point from Richard 🤔

osfameron (approved):
As a pure change, I think this is good to get the correction in.
Definitely worth making note of the overloaded term "buckets", and perhaps discussion on docs-style-and-tone and in future docs terminology discussions.

modules/android/pages/working-with-vector-search.adoc
ElliotFrancisHunter: I shall raise it with Iveta in the meantime, seems like a useful thing to discuss with Iveta.

PR #901, opened 2024-08-22
CBL-6039: Fix a mistake in Vector Search C# tagging and add more VS example

borrrden:


ElliotFrancisHunter (approved):
Looks great! Will integrate the snippets 👍🏼

PR #904, opened 2024-08-29
Update target JVM for JAK platforms

bmeike:
I see that there may be some references to `targetJvm`, `sourceCompatibility`, or targetCompatibility` in the text.  May want to check for those....


ElliotFrancisHunter (approved):
👍🏼

PR #908, opened 2024-09-03
DOC-11873: Further plat support changes

ElliotFrancisHunter:
Further changes to platform support and request to add back some items as requested by PM.
Preview Link:  https://preview.docs-test.couchbase.com/32platsupport/home/index.html

### DIFF
--- a/modules/ROOT/pages/_partials/supported-versions.adoc
+++ b/modules/ROOT/pages/_partials/supported-versions.adoc
@@ -46,7 +46,7 @@ Support for the following will be deprecated in this release and will be removed
 * macOS 11 - Big Sur
 * Apple OS X v10.14 (Mojave), v10.15 (Catalina)
 * Raspbian - 9
-* Debian 9
+* Debian 9, 10
 
 Please plan to migrate your apps to use an appropriate alternative version.
 --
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/supported-versions.adoc
```diff
@@ -325,7 +341,7 @@ Please plan to migrate your apps to use an appropriate alternative version.
 | OS|Version|Type
 
 |RedHat
-|9
+|8+
```
osfameron, line 344: ~wait what, Debian 8 is supported, but only 9,10 are deprecated?
Does that mean that 8 isn't deprecated?
Or that it was deprecated in a previous release so we don't have to mention it?~

Nope, misread. Sorry.

osfameron (approved):
approve to unblock ~but Q about Debian 8/9 inline.~

modules/ROOT/pages/_partials/supported-versions.adoc
```diff
@@ -46,7 +46,7 @@ Support for the following will be deprecated in this release and will be removed
 * macOS 11 - Big Sur
 * Apple OS X v10.14 (Mojave), v10.15 (Catalina)
 * Raspbian - 9
-* Debian 9
+* Debian 9, 10
```
bfavini, line 49: It's not the time to address this, but we should make sure that we are explicit about the releases here...is it 9.0 and 10.0? Or anything on either of these lines? Maybe a project to pursue when things settle down.

PR #909, opened 2024-09-04
DOC-12513: Fix android install instructions for vector search

ElliotFrancisHunter:
This PR fixes the missing android install instructions (The vector search activation), now that we have the missing snippet. 

Important to note is that the Kotlin snippet is the same as the Java one as Kotlin is able to parse certain Java functionality by design. (Had to look this up!)


modules/android/pages/gs-install.adoc
```diff
@@ -271,6 +272,13 @@ dependencies {
 }
 ----
```
RichardSmedley, line 275: There's a "+" character missing in the existing text, between lines 262 and 263,
which (in GH preview) is affecting the treatment of the ordered list, and causing your new content below to start the count again.

ElliotFrancisHunter: Fixed so ordering is displayed correctly.

RichardSmedley (approved):
LGTM

PR #910, opened 2024-09-13
DOC-13056: Antora assembler simplification

osfameron:
WIP with antora-assembler

Android database page looks pretty good to me, but haven't done full test!


RayOffiah (approved):
Haven't been through all the files, to be honest; but it looks like you've just removed all the `discrete` tags from the anchors.

So it should be good!

Ah! My mistake; got here a little too late.

PR #912, opened 2024-09-23
[BACKPORT] DOC-12568: Copy fix

ElliotFrancisHunter:
Backport PR for the customer found issue: https://jira.issues.couchbase.com/browse/DOC-12568

Fly-by grammar fix also.

### DIFF
--- a/modules/ROOT/pages/_partials/commons/common-sgw-replication.adoc
+++ b/modules/ROOT/pages/_partials/commons/common-sgw-replication.adoc
@@ -435,7 +435,7 @@ By default, when a user loses access to a channel, the next Couchbase Lite Pull 
 
 Documents that exist in multiple channels belonging to the user (even if they are not actively replicating that channel) are not auto-purged unless the user loses access to all channels.
 
-Users will be receive an `AccessRemoved` notification from the DocumentListener if they lose document access due to channel access revocation; this is sent regardless of the current auto-purge setting.
+Users will receive an `AccessRemoved` notification from the DocumentListener if they lose document access due to channel access revocation when the auto-purge setting is set to `true`.
 
 .Behavior following access revocation
 [#tbl-revoke-behavior, cols="^1h,2a,2a", options="header"]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/commons/common-sgw-replication.adoc
```diff
@@ -435,7 +435,7 @@ By default, when a user loses access to a channel, the next Couchbase Lite Pull
 
 Documents that exist in multiple channels belonging to the user (even if they are not actively replicating that channel) are not auto-purged unless the user loses access to all channels.
 
-Users will be receive an `AccessRemoved` notification from the DocumentListener if they lose document access due to channel access revocation; this is sent regardless of the current auto-purge setting.
+Users will receive an `AccessRemoved` notification from the DocumentListener if they lose document access due to channel access revocation when the auto-purge setting is set to `true`.
```
simon-dew: There's an ambiguity here: is it

1. "Users will receive (x if y) when z", or
2. "Users will receive x if (y when z)"

Do you see what I mean? I'm guessing it's the first one. To avoid this ambiguity, I would rephrase as

"When z, users will receive x if y."

```suggestion
When the auto-purge setting is set to `true`, users will receive an `AccessRemoved` notification from the DocumentListener if they lose document access due to channel access revocation.
```

simon-dew (dismissed):
One minor rephrase for clarity

simon-dew (approved):
LGTM 👍 thanks!

pasin:
@velicuvlad and @borrrden Can you look at the PR validation script? It seems to download the wrong binary.

pasin (approved):

borrrden:
The PR validation is meant to validate against in progress versions.  The jobs should be changed to run against 3.2.1 instead I think.

velicuvlad:
> The PR validation is meant to validate against in progress versions. The jobs should be changed to run against 3.2.1 instead I think.

That is correct.

I have an update for the iOS script anyway and will be contained in the upcoming PR that will address some snippet issues, so I'll fix the validation then, as it is not needed for this change.

velicuvlad (approved):

borrrden (approved):

ElliotFrancisHunter:
@borrrden @velicuvlad thanks for checking. I've now merged.

PR #915, opened 2024-09-26
DOC-12585: Upgrade notice admonition

ElliotFrancisHunter:
This is a PR for the escalated docs ticket: https://jira.issues.couchbase.com/browse/DOC-12585

Adding an downgrading section to minor release notes and the upgrade page in the docs for all platforms, the links below are the links for C.

Upgrade doc: https://preview.docs-test.couchbase.com/upgrade/couchbase-lite/current/c/upgrade.html

Release note: https://preview.docs-test.couchbase.com/upgrade/couchbase-lite/current/c/releasenotes.html

It states you cannot upgrade a 3.1.x database to 3.2.x and then downgrade again.


pasin (commented):
There is a paragraph that I think it might be for SG not CBL.

```
In addition, if you are syncing with a {version-full} Sync Gateway, you should be aware of the significant configuration enhancements introduced and their impact. This is a one-way conversion.
```

ElliotFrancisHunter:
> There is a paragraph that I think it might be for SG not CBL.
> 
> ```
> In addition, if you are syncing with a {version-full} Sync Gateway, you should be aware of the significant configuration enhancements introduced and their impact. This is a one-way conversion.
> ```

That is deliberate. It's a small signpost for users that are also making use of the Sync Gateway product. Have improved this by providing a reference to the page in question rather than just the statement.

RichardSmedley (approved):
:+1:

PR #917, opened 2024-09-27
CBL-6269: Update Getting Started page for CBL iOS

velicuvlad:


modules/objc/examples/code_snippets/GettingStarted.m
```diff
@@ -13,28 +22,31 @@ @implementation GettingStarted
...
     // Do any additional setup after loading the view.
-    // Get the database (and create it if it doesn’t exist).
-
-    bool *testReplication;
-    testReplication = false;
-
+    
+    [self getStartedWithReplication: false];
+    
+}
```
pasin, line 27: Better to have an empty line between methods.

modules/objc/examples/code_snippets/GettingStarted.m
```diff
@@ -13,28 +22,31 @@ @implementation GettingStarted
...
+- (void) getStartedWithReplication: (Boolean) replication {
```
pasin, line 29: What is this method for? Do we need any doc tags?

borrrden (commented):
Unsure what input I am supposed to provide

modules/objc/examples/code_snippets/GettingStarted.m
velicuvlad: for iOS no, the whole file is a snippet. All GettingStarted platforms specify that replication is OPTIONAL, although there is no optional in place for that piece of code in any other platform. Plan is to have another PR with those changes as it will be easier to review by the team - no specific iOS code in that one.

velicuvlad:
> Unsure what input I am supposed to provide

@borrrden Just FYI that .NET Jenkins will be fixed with this

pasin (approved):

ElliotFrancisHunter (approved):
All good on the docs front. Good job! 👍🏼

PR #919, opened 2024-10-10
Fix basic replication configuration C# snippet

borrrden:
Note the "description" numbered entries below this snippet also need some work.  They are no longer accurate, and don't actually line up correctly currently.


borrrden:
Also if you have a look at line 37 here https://mobile.jenkins.couchbase.com/blue/organizations/jenkins/_cbl_github_pipeline%2Fdocs-couchbase-lite/detail/PR-919/1/pipeline/10/#step-28-log-37 you can see a long list of warnings I made about either unused tags, or tags that are used but have no content.

PR #920, opened 2024-10-11
DOC-12636 simplify/update code samples for passive-peer csharp page

osfameron:
Using antora-assembler-simplify as starter.


borrrden (commented):
I don't think I understand what is happening here enough to make a review

PR #921, opened 2024-10-15
JAK GettingStarted: run the replicator only if there is a non-null SG URI

bmeike:
This will make it so that (potential) clients will be able to copy and paste a running app.

PR #922, opened 2024-10-17
CBL-6300: Update code on Getting Started docs page

velicuvlad:
- port 9b709f736156ab51911ad403aa44d9a9029648de from 3.2
- add Jenkins for iOS


modules/swift/examples/code_snippets/Getting-Started.swift
```diff
@@ -14,53 +24,36 @@ class ViewController: UIViewController {
...
-                print("Updated document id \(document.id), adding language \(document.string(forKey: "language")!)")
-            } catch {
-                fatalError("Error updating document")
-            }
-        }
+        mutableDoc = try collection.document(id: mutableDoc.id)!.toMutable()
+        mutableDoc.setString("Swift", forKey: "language")
+        try collection.save(document: mutableDoc)
+        let document = try collection.document(id: mutableDoc.id)
+        assert(document!.string(forKey: "language") == "Swift",
```
pasin: This is not the test so I don't think we should use assert here. Print info from the doc is fine.

modules/objc/examples/code_snippets/GettingStarted.m
```diff
@@ -43,27 +55,29 @@ - (void)viewDidLoad {
     // Create a query to fetch documents of type SDK.
     CBLQueryExpression *type = [[CBLQueryExpression property:@"type"] equalTo:[CBLQueryExpression string:@"SDK"]];
     CBLQuery *query = [CBLQueryBuilder select:@[[CBLQuerySelectResult all]]
-                                          from:[CBLQueryDataSource database:database]
+                                          from:[CBLQueryDataSource collection: collection]
```
pasin, line 58: I think we should stop showing QueryBuilder in the examples as it's not the maintenance mode now.

modules/objc/examples/code_snippets/GettingStarted.m
```diff
@@ -75,8 +89,9 @@ - (void)viewDidLoad {
         // Start replication
         [_replicator start];
 
+    } else {
+        NSLog(@"Not running replication");
```
pasin, line 93: I don't think this is important to be in the example?

modules/objc/examples/code_snippets/GettingStarted.m
```diff
@@ -43,27 +55,29 @@ - (void)viewDidLoad {
...
-                                          from:[CBLQueryDataSource database:database]
+                                          from:[CBLQueryDataSource collection: collection]
                                          where:type];
 
     // Run the query
     CBLQueryResultSet *result = [query execute:&error];
     NSLog(@"Number of rows :: %lu", (unsigned long)[[result allResults] count]);
 
-    if (testReplication) {
+    if (replication) {
```
pasin, line 65: Why we need this if here in the example code?

modules/objc/examples/code_snippets/GettingStarted.m
```diff
@@ -13,28 +22,31 @@ @implementation GettingStarted
...
     // Do any additional setup after loading the view.
-    // Get the database (and create it if it doesn’t exist).
-
-    bool *testReplication;
-    testReplication = false;
+    
+    [self getStartedWithReplication: false];
+}
 
+- (void) getStartedWithReplication: (Boolean) replication {
```
pasin, line 29: Is there a place in the doc that will call this with true or false?  It depends on the purpose of this big code snippet but I don't think we need this boolean flag at all?

pasin (approved):
The main thing I have in this PR is using the assertion and QueryBuilder in the code. If you can change that it will be great. 

However, in general, I don't understand the purpose of this big code snippet (Why we are doing a lot of things in one big code snippet? What are we trying to explain?)

modules/objc/examples/code_snippets/GettingStarted.m
velicuvlad: Yes, in `viewDidLoad`. The purpose of this is a longer discussion, but the main idea is that Repl with error on start due to the URL being dummy value if not changed. On top, there was no mention for iOS that SGW is needed. Added to that, all platforms have code comment that replication is optional, which is not reflected in the "implementation". I believe the main purpose of a such called `GettingStarted` is to verify that the CBL installation can actually run some basic operations. As we are going to have a lot more new developers, I believe this code might be copy-pasted-run so it should error/crash.

I've asked Blake to make similar changes https://github.com/couchbase/docs-couchbase-lite/pull/921 . For .NET and C, as we have validation pipeline in place, I thought I will give a try by myself and have you and Jim review afterwards (only for 3.2 docs).

modules/objc/examples/code_snippets/GettingStarted.m
velicuvlad: explained below -> to guard against copy-paste-run and what not because repl errors on run as the URL endpoint is a placeholder and we dont instruct anywhere that that should be changed. An if statement is language agnostic and easy to view/read.

modules/objc/examples/code_snippets/GettingStarted.m
velicuvlad: This is to make the user look at the replication code and see that is optional, which will also point out the URL

modules/objc/examples/code_snippets/GettingStarted.m
velicuvlad: this PR is with 3.1. I would keep this here as is, thinking we do not plan to release another dozen of 3.1s. I will make myself a note to strip this from 3.2.

modules/swift/examples/code_snippets/Getting-Started.swift
velicuvlad: replaced with log. Will note this as well to fix in 3.2 line as I checked and is there as well

velicuvlad:
@ElliotFrancisHunter can you please have another sanity check for this on 3.1 as well?

ElliotFrancisHunter:
> The main thing I have in this PR is using the assertion and QueryBuilder in the code. If you can change that it will be great.
> 
> However, in general, I don't understand the purpose of this big code snippet (Why we are doing a lot of things in one big code snippet? What are we trying to explain?)

@pasin I think the idea of the big snippet is it's a basic, self-contained app allowing users to quickly get started with CBL on a given platform, demonstrating basic CRUD operations, simple query and optional Sync Gateway replication. It allows them to copy the whole thing across into their application to get started.

ElliotFrancisHunter (approved):
Looks good from a docs POV. Approved 👍🏼

PR #932, opened 2024-11-19
DOC-12747: Consistent formatting

ElliotFrancisHunter:
Reset indent across platforms so they look the same. .NET screenshot attached for reference.
Screenshot 2024-11-19 at 11 47 50


RichardSmedley (approved):
:+1:

PR #934, opened 2024-11-22
Doc 12623 logging version fix

ElliotFrancisHunter:

Underlying cause is the cbl log tool was only released every minor/major version so the wget command would always fail using the `version-full` attribute once a patch version was involved.


RichardSmedley (approved):
:+1:

PR #935, opened 2024-11-22
DOC-12623: Explicitly define log version

ElliotFrancisHunter:
This is required as the cbl-log tool only has releases per minor/major version and NOT maintenance version. Rendering examples incorrect.

Underlying cause is the cbl log tool was only released every minor/major version(latest being 3.0.0) so the wget command would always fail using the version-full attribute once a patch version was involved.


RichardSmedley (approved):
:+1:

PR #936, opened 2024-11-22
DOC-12623: Correct logging version.

ElliotFrancisHunter:
Enforce 3.0.0 as it's the latest version. Also add an admonition to clarify to customers.

Underlying cause is the cbl log tool was only released every minor/major version (latest being 3.0.0) so the wget command would always fail using the version-full attribute once a patch version was involved.


RichardSmedley (approved):
:+1:

PR #937, opened 2024-11-22
DOC-12623: Fix logging version.

ElliotFrancisHunter:
Enforce cbl log tool version 3.0.0
The tool had previously been updated every minor/major release but stopped at 3.0. No functionality changes to the tool have occurred for some time. For now, users can correctly download version 3.0 if required.

Underlying cause is the cbl log tool was only released every minor/major version so the wget command would always fail using the version-full attribute once a patch version was involved.


ElliotFrancisHunter:
Once again, ignore build error as it has been fixed in another PR and is related to snippets rather than doc.

RichardSmedley (approved):
:+1:

PR #945, opened 2025-02-04
DOC-12925: Logging API snippets iOS

velicuvlad:
- New snippets for the new Logging API 3.2.2
- Cleaned up a bit the existing tags so there is no duplicate code and to respect the ticket description. We will need to verify where those were used and replace them, if needed.

### DIFF
--- a/modules/objc/examples/code_snippets/SampleCodeTest.m
+++ b/modules/objc/examples/code_snippets/SampleCodeTest.m
@@ -211,19 +211,6 @@
 
 #pragma mark - Logging
 
-- (void) dontTestLogging {
-    // tag::logging[]
-
-    // Replicator / Verbose
-    CBLDatabase.log.console.level = kCBLLogLevelVerbose;
-    CBLDatabase.log.console.domains = kCBLLogDomainReplicator;
-
-    // Query /  Verbose
-    CBLDatabase.log.console.level = kCBLLogLevelVerbose;
-    CBLDatabase.log.console.domains = kCBLLogDomainQuery;
-    // end::logging[]
-}
-
 #if COUCHBASE_ENTERPRISE
 - (void) dontTestDatabaseEncryption {
     // tag::database-encryption[]
--- a/modules/objc/examples/code_snippets/SampleCodeTest.m
+++ b/modules/objc/examples/code_snippets/SampleCodeTest.m
@@ -242,8 +242,8 @@
     // tag::file-logging[]
     NSString *tempFolder = [NSTemporaryDirectory() stringByAppendingPathComponent: @"cbllog"];
     CBLLogFileConfiguration *config = [[CBLLogFileConfiguration alloc] initWithDirectory:tempFolder]; // <.>
-    config.maxRotateCount = 2; // <.>
-    config.maxSize = 1024; // <.>
+    config.maxRotateCount = 12; // <.>
+    config.maxSize = 524288; // <.>
     config.usePlainText = YES; // <.>
     [CBLDatabase.log.file setConfig:config];
     [CBLDatabase.log.file setLevel:kCBLLogLevelVerbose]; // <.>
@@ -256,8 +256,8 @@
     CBLLogSinks.file = [[CBLFileLogSink alloc] initWithLevel:kCBLLogLevelVerbose
                                                    directory:tempFolder
                                                 usePlaintext:false
-                                                maxKeptFiles:2
-                                                 maxFileSize:1024];
+                                                maxKeptFiles:12
+                                                 maxFileSize:524288];
     // end::new-file-logging[]
 }
 
--- a/modules/objc/examples/code_snippets/SampleCodeTest.m
+++ b/modules/objc/examples/code_snippets/SampleCodeTest.m
@@ -102,6 +102,21 @@
 
 // end::custom-logging[]
 
+// tag::new-custom-logging[]
+@interface TestLogger :NSObject<CBLLogSinkProtocol>
+
+@end
+
+@implementation TestLogger
+
+- (void) writeLogWithLevel:(CBLLogLevel)level domain:(CBLLogDomain)domain message:(NSString*)message {
+    // handle the message, for example piping it to
+    // a third party framework
+}
+
+@end
+// end::new-custom-logging[]
+
 // tag::local-win-conflict-resolver[]
 @interface LocalWinConflictResolver :NSObject<CBLConflictResolver>
 @end
@@ -194,6 +209,7 @@
   // end::database-fullsync[]
 }
 
+#pragma mark - Logging
 
 - (void) dontTestLogging {
     // tag::logging[]
@@ -226,13 +242,14 @@
     // tag::console-logging[]
     CBLDatabase.log.console.domains = kCBLLogDomainAll; // <.>
     CBLDatabase.log.console.level = kCBLLogLevelVerbose; // <.>
-
     // end::console-logging[]
+}
 
-    // tag::console-logging-db[]
-    CBLDatabase.log.console.domains = kCBLLogDomainAll;
-
-    // end::console-logging-db[]
+- (void) dontTestEnableNewConsoleLogging {
+    // tag::new-console-logging[]
+    CBLConsoleLogSink* sink = [[CBLConsoleLogSink alloc] initWithLevel:kCBLLogLevelVerbose domains:kCBLLogDomainAll];
+    CBLLogSinks.console = sink;
+    // end::new-console-logging[]
 }
 
 - (void) dontTestFileLogging {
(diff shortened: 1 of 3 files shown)


modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -226,13 +229,14 @@ - (void) dontTestEnableConsoleLogging {
...
+}
 
-    // tag::console-logging-db[]
-    CBLDatabase.log.console.domains = kCBLLogDomainAll;
-
-    // end::console-logging-db[]
+- (void) dontTestEnableNewConsoleLogging {
+    // tag::new-console-logging[]
+    CBLConsoleLogSink* sink = [[CBLConsoleLogSink alloc] initWithLevel:kCBLLogLevelVerbose domains:kCBLLogDomainAll];
+    CBLLogSinks.console = sink;
```
pasin: Could be in one line? (Same of the rest of the code snippets in this PR).

```
CBLLogSinks.console = [[CBLConsoleLogSink alloc] initWithLevel: kCBLLogLevelVerbose domains: kCBLLogDomainAll];
```

Also, it's a good idea  to consistently use the space or not using the pace after `:`.

modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -102,6 +102,21 @@ - (void) logWithLevel:(CBLLogLevel)level domain:(CBLLogDomain)domain message:(NS
 
 // end::custom-logging[]
 
+// tag::new-custom-logging[]
+@interface TestLogger :NSObject<CBLLogSinkProtocol>
```
pasin: TestLogSink to be inline with CBLLogSinkProtocol?

modules/objc/examples/code_snippets/SampleCodeTest.m
```diff
@@ -247,13 +250,26 @@ - (void) dontTestFileLogging {
     // end::file-logging[]
 }
 
+- (void) dontTestNewFileLogging {
+    // tag::new-file-logging[]
+    NSString* tempFolder = [NSTemporaryDirectory() stringByAppendingPathComponent:  @"cbllog"];
+    CBLLogSinks.file = [[CBLFileLogSink alloc] initWithLevel:kCBLLogLevelInfo directory:tempFolder];;
```
pasin: double `;;`

pasin (approved):
There is a double ';;'. Please check.

modules/swift/examples/code_snippets/SampleCodeTest.swift
```diff
@@ -120,12 +108,25 @@ class SampleCodeTest {
         Database.log.file.level = .info // <.>
         // end::file-logging[]
     }
+    
+    func dontTestNewFileLogging() throws {
+        // tag::new-file-logging[]
+        let tempFolder = NSTemporaryDirectory().appending("cbllog")
+        LogSinks.file = FileLogSink(level: .info, directory: tempFolder)
```
pasin: Just noticed this from Java Snippet. It is a good idea to be consistent.

```
LogSinks.get().setFile(new FileLogSink.Builder()
            .setDirectory("/tmp/logs")
            .setMaxKeptFiles(12)
            .setPlainText(false)
            .build());
```

modules/swift/examples/code_snippets/SampleCodeTest.swift
```diff
@@ -120,12 +108,25 @@ class SampleCodeTest {
...
-        // tag::set-custom-logging[]
+        // tag::custom-logging[]
         let logger = LogTestLogger(.warning)
         Database.log.custom =  logger // <.>
-        // end::set-custom-logging[]
+        // end::custom-logging[]
+    }
+    
+    func dontTestNewCustomLogging() throws {
+        // tag::new-custom-logging[]
```
pasin, line 120: There are two `new-custom-logging` tags. Will this work? Probably, this one need to be `set-new-custom-logging[]`?

modules/swift/examples/code_snippets/SampleCodeTest.swift
```diff
@@ -85,29 +85,17 @@ class SampleCodeTest {
...
-        // tag::console-logging-db[]
-
-        Database.log.console.domains = .database
-
-        // end::console-logging-db[]
+    }
+    
+    func dontTestNewConsoleLogging() throws {
+        // tag::new-console-logging[]
+        LogSinks.console = ConsoleLogSink(level: .verbose, domains: .all)
```
pasin: domains is optional and normally we want users to set it to all. Java doesn't include domains. So I think it is a good idea to just not include the domains parameter here.

modules/objc/examples/code_snippets/SampleCodeTest.m
pasin: I think it's a good idea to set he log level to verbose as that is what we normally want users to do.

modules/swift/examples/code_snippets/SampleCodeTest.swift
pasin: I also think it's a good idea to set the log level to verbose as that is what we want users to do.

modules/swift/examples/code_snippets/SampleCodeTest.swift
ElliotFrancisHunter: Is there any specific benefit the `domains` parameter provides on iOS? It may be useful to call out in docs if so.

pasin: The benefit of the domains is mostly for debugging in a specific area. For example, if you want to debug an issue related to query, you can specify only the `query` domain to filter out the logs that are not related to query.

pasin: It's fine to leave the domains parameter there too if you think that is better.

velicuvlad: Until we remove the existing (old) API from docs, I am going to match whatever was there before for it.

Java snippet uses the full constructor for FileSink, so I will change to do the same. Thus, I will keep domains here even though is optional, it is matching the older snippet.

modules/swift/examples/code_snippets/SampleCodeTest.swift
velicuvlad: I have raised this to @ElliotFrancisHunter offline. My understanding is that it should work. If not, I will add back the new tag `set-new-custom-logging[]`.

modules/swift/examples/code_snippets/SampleCodeTest.swift
```diff
@@ -103,16 +101,17 @@ class SampleCodeTest {
...
-        Database.log.file.level = .info // <.>
+        Database.log.file.level = .verbose // <.>
         // end::file-logging[]
     }
     
     func dontTestNewFileLogging() throws {
         // tag::new-file-logging[]
         let tempFolder = NSTemporaryDirectory().appending("cbllog")
-        LogSinks.file = FileLogSink(level: .info, directory: tempFolder)
+        LogSinks.file = FileLogSink(level: .verbose, directory: tempFolder, usePlainText: false, maxKeptFiles: 2, maxFileSize: 1024)
```
pasin: This is not a good example that will not generally work well and a lot of log will be missing as the number of files keep and max file size are too low.

Something like this will be better and consistent with Java. The default max file size is not too bad.

```
LogSinks.file = FileLogSink(level: .verbose, directory: tempFolder, usePlainText: false, maxKeptFiles: 12)
```

pasin (dismissed):

velicuvlad:
Re-arranged around the snippets as I really liked the way Jim organised it for .NET.

modules/swift/examples/code_snippets/SampleCodeTest.swift
ElliotFrancisHunter: @velicuvlad I've reviewed the code. In _theory_ this should be fine, but it means that the tagged regions will be included together as one unit. I think separating into `set-new-custom-logging` and `new-custom-logging` will be safer. Can we change to that please?

ElliotFrancisHunter (changes requested):
As I said in my comment, the extra tag name is required before I can approve and merge this.

modules/swift/examples/code_snippets/SampleCodeTest.swift
velicuvlad: @pasin  @ElliotFrancisHunter they should be included as 1 region. We want both of those snippets together as you need to define a logger that obeys SinkProtocol (writeLog) before plugging it into the CustomLogSink config. 

I can still change it back, but my opinion/understanding is that these 2 should be together. (I also believe there is no place where this should be separated, hence no need for extra tag)

ElliotFrancisHunter: @pasin @velicuvlad my question then would be why is there a need to separate them in the code rather than one block with the whole functionality?

velicuvlad: To be able to build the project/certify the snippets. I can't declare/implement a class inside of another class. Can be done for Swift, but Obj-C doesn't allow for nested classes. It is easier to follow if both are the same.

ElliotFrancisHunter (approved):
Happy to merge after our conversation 👍🏼

PR #947, opened 2025-02-06
CBL-6683 : C - Partial Index Snippets

pasin:
* Added Partial Index Code Snippet : partial-value-index and partial-full-text-index.

* Updated the other index snippet to show the parameter names.

### DIFF
--- a/modules/c/examples/code_snippets/main.cpp
+++ b/modules/c/examples/code_snippets/main.cpp
@@ -1027,10 +1027,9 @@ static void create_index() {
     // Syntax for second argument is the same as taking from a N1QL SELECT
     // i.e. SELECT (type, name) FROM _;
     // tag::scopes-manage-index-collection[]
-    CBLValueIndexConfiguration config = {
-        kCBLN1QLLanguage,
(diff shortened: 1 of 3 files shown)


modules/c/examples/code_snippets/main.cpp
```diff
@@ -1067,17 +1064,46 @@ static void create_array_index_nested() {
...
+
+static void create_partial_full_text_index() {
+    CBLDatabase* database = kDatabase;
+    CBLCollection* collection = CBLDatabase_DefaultCollection(kDatabase, NULL);
+
+    // tag::partial-full-text-index[]
+    CBLFullTextIndexConfiguration config{};
+    config.expressionLanguage = kCBLN1QLLanguage;
+    config.expressions = FLSTR("content");
+    config.where = FLSTR("year > 1999");
```
velicuvlad, line 1100: Made a comment in the ticket related to this - also see iOS PR. It doesn't make much of a difference, but would prefer if we all can have the same (no matter which one). Otherwise, LGTM.

velicuvlad (approved):

PR #948, opened 2025-02-07
CBL-6748: Add new logging API snippets

borrrden:
Added two more tags since to demonstrate custom logging you need to show the class being used

### DIFF
--- a/modules/csharp/examples/code_snippets/Program.cs
+++ b/modules/csharp/examples/code_snippets/Program.cs
@@ -1804,7 +1804,7 @@ namespace api_walkthrough
             // end::tojson-blob[]
         }
 
-        private void CreateArrayIndex()
+        public void CreateArrayIndex()
         {
             var database = new Database("my-database");
             var collection = database.GetDefaultCollection();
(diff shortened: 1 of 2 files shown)


modules/csharp/examples/code_snippets/Program.cs
```diff
@@ -2602,4 +2602,72 @@ public void InitReplication()
...
+    public void NewLoggingApi()
+    {
+        // tag::new-console-logging[]
+        LogSinks.Console = new ConsoleLogSink(LogLevel.Verbose);
+        // end::new-console-logging[]
+
+        // tag::new-file-logging[]
+        LogSinks.File = new FileLogSink(LogLevel.Verbose, "path/to/log/directory")
+        {
+            MaxKeptFiles = 3, // Save 3 log files (i.e. 2 rotated and 1 current)
```
velicuvlad, line 2664: I haven't checked the .NET implementation. But isn't this supposed to be the same value for the user (2)? I know we accounted for the -/+ 1 when passing the param around (at least I do for iOS).

borrrden: Handling the offset by one from core happens underneath the API but it doesn’t make sense to name the property “max kept files” and not have it describe how many files exist. You can check the defaults which are 1 for the old api and 2 for the new.

ElliotFrancisHunter (approved):

PR #951, opened 2025-02-14
DOC-12950 CBL 3.2.2 doc

ElliotFrancisHunter:
### Release Notes Reference:
https://confluence.issues.couchbase.com/wiki/spaces/cbeng/pages/2770305028/CBL3.2.2ReleaseNotes

## Changed Pages
-------------------

- What's new: https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/cbl-whatsnew.html#release-3-2-2-feburary-2025

- Release Notes (Android): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/releasenotes.html
- Release Notes (C): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/c/releasenotes.html
- Release Notes (.NET): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/csharp/releasenotes.html
- Release Notes (Java): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/java/releasenotes.html
- Release Notes (Objective-C): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/objc/releasenotes.html
- Release Notes (Swift): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/swift/releasenotes.html

- New Logging API (Android): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/new-logging-api.html
- New Logging API (C): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/c/new-logging-api.html
- New Logging API (.NET): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/csharp/new-logging-api.html
- New Logging API (Java): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/java/new-logging-api.html
- New Logging API (Objective-C): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/objc/new-logging-api.html
- New Logging API (Swift): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/swift/new-logging-api.html

- Partial Index (Android): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/indexing.html#partial-index
- Partial Index (C): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/c/indexing.html#partial-index
- Partial Index (.NET): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/csharp/indexing.html#partial-index
- Partial Index (Java): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/java/indexing.html#partial-index
- Partial Index (Objective-C): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/objc/indexing.html#partial-index
- Partial Index (Swift): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/swift/indexing.html#partial-index

- Old Logging API deprecation Notice (Android): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/troubleshooting-logs.html#introduction
- Old Logging API deprecation Notice (.NET): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/csharp/troubleshooting-logs.html#introduction
- Old Logging API deprecation Notice (Java): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/java/troubleshooting-logs.html#introduction
- Old Logging API deprecation Notice (Objective-C): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/objc/troubleshooting-logs.html#introduction
- Old Logging API deprecation Notice (Swift): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/swift/troubleshooting-logs.html#introduction

----------------------
## Pages still needing changes/updates
-----------------------

....

-----------------------


modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,84 @@
...
+You can find information about the new Couchbase Lite Logging API introduced in Couchbase Lite 3.2.2.
+For information about the now deprecated earlier version of the Logging API, see xref:3.1@couchbase-lite:{param-module}:troubleshooting-logs.adoc[]
+
+endif::is-c[]
+
+== LogSinks
+
+A lot of the new logging API will seem very familiar: many of the concepts are
+the same.  There are still three places that log messages can be sent: the log
+files, the console, and custom code and Couchbase still logs its messages in a
```
RayOffiah:
```suggestion
files, the console, and custom code. Couchbase still logs its messages in a
```

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,84 @@
...
+handfull of named "domains" and at common log levels: `debug` the most verbose,
```
RayOffiah:
```suggestion
handful of named "domains" and at common log levels: `debug` the most verbose,
```

modules/ROOT/pages/_partials/commons/common-indexing.adoc
```diff
@@ -109,6 +109,215 @@ include::{root-partials}block_tabbed_code_example.adoc[]
...
+SELECT * FROM tab2 WHERE b=456 AND c<>0;  -- uses partial index
+
+---- 
+
+Whereas the next query can not use the partial index:
+
+[source, sqlpp]
+----
+
+SELECT * FROM tab2 WHERE b=456;  -- cannot use partial index
```
RayOffiah: Quick question:
This cannot use a partial index; is that because  the query isn't AND-connected?

RayOffiah (dismissed):
Two small changes and a question to satisfy my curiosity.

modules/ROOT/pages/_partials/commons/common-indexing.adoc
ElliotFrancisHunter: Pretty much! The index created in the snippet below for reference:

`CREATE INDEX ex2 ON tab2(b,c) WHERE c IS NOT NULL;`

The key thing here is WHERE c IS NOT NULL. The second query is related to the IMPORTANT admonition. It requires satisfaction of the first rule which is fails. The WHERE clause is the created index `X` is not AND-connected and the WHERE clause in the query `W` is not OR-connected and contains none of the terms from the WHERE clause - as in it is missing reference to `c`

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
bmeike: This change is grammatically incorrect.  Adding the "and" makes it seem that there are four log sinks.  There are three: file, console and custom.  The next sentence describes the first of these, the file log sink.  Perhaps it should begin a new paragraph?

bmeike (commented):
I do not think replacing the period with a conjunction is a good idea.

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
ElliotFrancisHunter: Done!

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
ElliotFrancisHunter: Resolving as stale

modules/ROOT/pages/cbl-whatsnew.adoc
```diff
@@ -17,6 +17,82 @@ NOTE: Couchbase Lite 3.0 introduces some breaking changes. +
...
+==== New Logging API
+
+Couchbase Lite 3.2.2 introduces a new Logging API.
+The new Logging API has the following benefits:
+
+* Backwards compatible with earlier version of Couchbase Lite.
+
+* Removed polling overhead on custom loggers.
+
+* Simplified platform code and implementation.
```
idulo: this is more of a benefit for us, maybe for users: simplified API and reduced implementation complexity? let's put it last in the list as overall benefit

bmeike:
Here: New Logging API (Android): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/new-logging-api.html
Only the last point is of any interest to the user.
The new API is, absolutely, not backwards compatible.  It is, internally, backwards compatible: the code we wrote will support the old API while it is deprecated but not removed.  For our users, it is completely incompatible.

bmeike:
Here: New Logging API (Java): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/java/new-logging-api.html
"Here logs are written to separate log files filtered by log level, with each log level supporting individual retention policies."
This is not true.  Each log level is written to a separate file and there is a single retention policy for all files.

bmeike:
Here: Old Logging API deprecation Notice (Android): https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/troubleshooting-logs.html#introduction
"setRotateCount from the old API is slightly different from setMaxKeptFiles. setRotateCount is the maximum size in bytes of a log file. The latter is the maximum number of log files that will exist at any time and is the count of rotated files plus one. "

This is not true.  `setRotateCount` has nothing to do with the size of the log file.  It is the number of rotated files. It is maxKeptFiles - 1

bmeike:
Umm... This is just completely wrong: https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/java/troubleshooting-logs.html#lbl-custom-logs
This is the old API.  For the new API, you need to `extend` (*not* `implement`) `BaseLogSink`
The example snippets are tagged with:  `tag::new-custom-logging[]`
 ```
   public void newCustomLoggingExample(Fn.Consumer<String> sendToNetwork) {
        // tag::new-custom-logging[]
        LogSinks.get().setCustom(new BaseLogSink(LogLevel.WARNING, LogDomain.NETWORK, LogDomain.REPLICATOR) {
            @Override
            public void writeLog(LogLevel level, LogDomain domain, String message) {
                // this method will be called only with messages from the NETWORK and REPLICATOR
                // domains with a log level of WARNING or higher.
                sendToNetwork(String.format("%s/%s: %s", domain, level, message));
            }
        });
        // end::new-custom-logging[]
```

**Review by bmeike — DISMISSED** [review]
There are several miss-communications here that should be corrected
I've commented them.  Holler if any of the comments are unclear.

File: modules/ROOT/pages/cbl-whatsnew.adoc (hunk 4 above)
Comment (ElliotFrancisHunter) [reply modules/ROOT/pages/cbl-whatsnew.adoc]: Done!

**Review by osfameron — APPROVED** [review]

**Comment by bmeike** [comment]
On this page:
https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/new-logging-api.html

== This:
"at common log levels: debug the most verbose, and error only for serious failures." should probably read:
 at common log levels: LogLevel.DEBUG the most verbose, and LogLevel.ERROR only for serious failures.
 
 
== This:
The new Logging API has the following benefits:
    Removed polling overhead on custom loggers.
    You can enque Log callbacks from LiteCore for asynchronous delivery.
    Log sinks are now thread safe, removing risk of inconsistent states during initialization.
    Simplified API and reduced implementation complexity.

Could be simplified.  Only the last two points are visible to client code.  The second point is probably incorrect....

== https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/new-logging-api.html#lbl-console-logsink
Perhaps those two logging levels should be the same?

== https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/android/new-logging-api.html#lbl-custom-logsink
I suggest removing this sentence entirely:
"Unless otherwise specified logs from all Couchbase domains are forwarded to the custom log sink."
The sentence below it is correct and this sentence appears to contradict it.

**Comment by bmeike** [comment]
== https://preview.docs-test.couchbase.com/322Doc/couchbase-lite/current/java/new-logging-api.html#lbl-file-logsink
I think the note should read:
setRotateCount from the old API is slightly different from setMaxKeptFiles. The latter is the maximum number of log files that will exist at any time and is the count of rotated files (setRotateCount) plus one.

**Comment by bmeike** [comment]
@ElliotFrancisHunter This looks great.  Awesome.

I'm not an expert on Partial Indexing: parsing and evaluating those queries happens entirely in LiteCore. If, maybe, Pasin approves the docs for any one platform, I think they should be right for all.  I note that you have the updated code samples, so thats good.

I still have a couple of suggestions that I think will help avoid confusion, but I no longer see anything I think is incorrect.

**Review by bmeike — APPROVED** [review]
I have a couple suggestions, but I don't see anything blocking

File: modules/ROOT/pages/_partials/commons/common-indexing.adoc [hunk 5]
```diff
@@ -109,6 +109,215 @@ include::{root-partials}block_tabbed_code_example.adoc[]
...
+The operators include `=, ==, <, >, <=, >=< <>, IN, LIKE, and BETWEEN`.
+
+IMPORTANT: If `X` is in the form of `"z is NOT NULL"` or `"z is VALUED"`, the first rule must be satisfied.
+
+For example, let the index be:
+
+[source, sqlpp]
+----
+
+CREATE INDEX ex2 ON tab2(b,c) WHERE c IS NOT NULL;
```
pasin: CBL doesn't support this create index syntax and we are using the API for creating the index. For simplicity, I think we can just say the following without using the actual code snippet:

```
For example, let the partial index be `c IS NOT NULL` when creating a partial index on collection named "col1" in the default scope.
```

modules/ROOT/pages/_partials/commons/common-indexing.adoc
```diff
@@ -109,6 +109,215 @@ include::{root-partials}block_tabbed_code_example.adoc[]
...
+----
+
+Then any query that uses operators `=, <, >, <=, >=, <>, IN, LIKE, or GLOB` on column `c` would be usable with the partial index, because those comparison operators are only true if `c` is not NULL. 
+
+The following query could use the partial index:
+
+[source, sqlpp]
+----
+
+SELECT * FROM tab2 WHERE b=456 AND c<>0;  -- uses partial index
```
pasin: I suggested to use "col1" above as tab is more like a table name.

modules/ROOT/pages/_partials/commons/common-indexing.adoc
pasin: I suggested to use "col1" above as tab is more like a table name.

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,634 @@
...
+
+== LogSinks
+
+Couchbase Lite 3.2.2 introduces a new Logging API.
+
+The first thing to note is that the three destinations for logs have been renamed as `LogSinks`, in keeping with common source/sink terminology. 
+
+The `FileLogSink`, the `ConsoleLogSink` and the `CustomLogSink` are all installed in an instance of `LogSinks`. 
+
+All three log sinks are initialized by default.
```
pasin: It seems this could give an impression that all log sinks are enabled. I think it's more accurate to say that only ConsoleLogSink is enabled for all logging domains at the warning log level by default.

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,634 @@
...
+The first thing to note is that the three destinations for logs have been renamed as `LogSinks`, in keeping with common source/sink terminology. 
+
+The `FileLogSink`, the `ConsoleLogSink` and the `CustomLogSink` are all installed in an instance of `LogSinks`. 
+
+All three log sinks are initialized by default. 
+If you want to stop all logging to one of them, though, you can set the sink to null.
+
+For example, you can stop logging to the FileLogSink like this:
+
+`LogSinks.get().setFile(null)`
```
pasin: This is very specific to Java / Android and as only console log sink is enabled by default, I think we could just say something like :

```
To enable a specific type of LogSink, create a LogSink object of that type, set its minimum log level and domains, and assign it to LogSinks. To disable a LogSink, set it to null or use a LogSink with LogLevel.NONE.
```

The code snippet for disabling the log sink should be shown here, we will need the code snippet for each platform. Maybe the current code snippet for setting each type of log sink is enough?

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,634 @@
...
+
+Couchbase still logs its messages in a handful of named domains and at common log levels: `LogLevel.DEBUG` the most verbose, and `LogLevel.ERROR` only for serious failures.
+
+To get the instance of `LogsSinks`, use the method `LogSinks.get()`.
```
pasin: `LogSinks.get()` is Java/Android specific and it should be `LogSinks` not `LogsSinks`. 

Maybe, `Accessing LogSinks depends on the platform. In Java, use LogSinks.get() to obtain an instance. In Swift, Objective-C, .NET, and C, LogSinks provides only static methods, without a singleton instance.`

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,634 @@
...
+include::c:example$code_snippets/main.cpp[tags=new-file-logging]
+
+----
+endif::is-c[]
+
+[#lbl-custom-logsink]
+=== Using a Custom Logger
+
+Installing a custom log sink with the new API is also streamlined: create an
+instance of your custom sink and use `LogSinks.get().setCustom` to install it.
```
pasin: Maybe `Use LogSinks.get().setCustom in Java or the corresponding static methods on other platforms.`

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,634 @@
...
+
+As with the other log sinks, you will have to specify the level and domain
+at which Couchbase logs are forwarded to your custom sink at its creation.
+
+Your custom log sink code will have to change as well.  The most significant
+change is that instead of _implementing_ the `Logger` interface, a custom log sink
+must _extend_ the `BaseLogSink` class.  
+
+The `BaseLogSink` class does not have a no-args constructor: you must specify at least the level at which Couchbase logs will be forwarded to the logger.
```
pasin: Not sure what is the best way to describe this in the cross-platform way. The `BaseLogSink` described here only applies to Java/Android and .NET. 

In Swift and Objective-C, which lack abstract methods, a custom log sink implements LogSinkProtocol and is assigned to a CustomLogSink instance during creation.

Similarly, in C, the custom log sink implementation is a callback that is assigned to CBLCustomLogSink object during creation.

pasin (changes requested):

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
velicuvlad: I second this. Moreover, I think it's "more correct" to maybe say that "all are to be installed in the LogSinks instance". There is only 1 instance of LogSinks and is available to the user right away.

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
velicuvlad: For the generated previews, this is the case for iOS as well - the Java specific snippet. More specifically, Swift should be `LogSinks.file = nil` instead. Similar for Obj-C `CBLLogSinks.file = nil`. And so, if we don't go the route to be adding additional snippets for invalidating sinks (disable logging), Pasin's suggestion is good/universal for all platforms.

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,634 @@
...
+The `BaseLogSink` class does not have a no-args constructor: you must specify at least the level at which Couchbase logs will be forwarded to the logger.  
+
+A second important change is that your logger will receive only logs at the level and domain for which it is initialized. 
+There is no need to record or filter the logs forwarded to the protected `writeLog` method which replaces the public `log` method from the old API.
+
+Related to this last point, the Couchbase `Loggers`, now `LogSinks` are meant to support logging by the Couchbase Lite platform.
+They were never meant as a general framework for logging. 
+
+IMPORTANT: With the new API, customer code can no longer log, directly, to any of the Couchbase log sinks.  
+The Console and File logsinks are final classes and do not publish methods that allow writing logs.
```
velicuvlad: _logsinks are final classes_ 

Although, from the user perspective, both a final class and a struct cant be subclassed, this is not true for iOS. Swift LogSinks are structs and Obj-C doesn't have a final class concept (we "manually" prevent against subclassing.)

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
velicuvlad: I think that for the custom log sink every platform should have its own rubric with a brief explanation and the specifics. There are snippets from old to new api just below that are platform specific as well and so it should suffice.

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
```diff
@@ -0,0 +1,634 @@
...
+include::swift:example$code_snippets/SampleCodeTest.swift[tags=file-logging]
+
+----
+
+.New API
+:!example-caption:
+[source, swift]
+----
+
+include::swift:example$code_snippets/SampleCodeTest.swift[tags=new-file-logging]
```
velicuvlad, line 355: I need to make an update to this - to add `maxSize`, such that is matching 1to1 with the old snippet.

modules/csharp/pages/new-logging-api.adoc
```diff
@@ -0,0 +1,18 @@
...
+:keywords: edge mobile api java kotlin android apple logging file console custom 
+
+[abstract]
+{description}
+
+// Sets the correct mobile sdk for xref links in the common page
+include::partial$_set_page_context_for_csharp.adoc[]
+
+// ROOT:partial$commons/common-
+include::{root-commons}new-logging-api.adoc[]
```
borrrden, line 18: This partial contains Java style code snippets inline

modules/ROOT/pages/_partials/commons/common-indexing.adoc
```diff
@@ -109,6 +109,215 @@ include::{root-partials}block_tabbed_code_example.adoc[]
...
+include::java:example$snippets/common/main/java/com/couchbase/codesnippets/QueryExamples.java[tags=query-partial-index]
+
+----
+
+endif::is-java[]
+ifdef::is-csharp[]
+
+[source, csharp, indent=0]
+----
+include::csharp:example$code_snippets/Program.cs[tags=array-index-config]
```
borrrden: Why is C# only using a different tag than the others?

modules/ROOT/pages/_partials/commons/common-indexing.adoc
```diff
@@ -109,6 +109,215 @@ include::{root-partials}block_tabbed_code_example.adoc[]
...
+include::java:example$snippets/common/main/java/com/couchbase/codesnippets/QueryExamples.java[tags=query-partial-full-index]
+
+----
+
+endif::is-java[]
+ifdef::is-csharp[]
+
+[source, csharp, indent=0]
+----
+include::csharp:example$code_snippets/Program.cs[tags=array-index-config]
```
borrrden: Here C# is also using a different tag, this seems incorrect

borrrden (dismissed):
Also agree with what Pasin has said.

modules/ROOT/pages/_partials/commons/common-indexing.adoc
simon-dew: Done

modules/ROOT/pages/_partials/commons/common-indexing.adoc
simon-dew: Done

modules/ROOT/pages/_partials/commons/common-indexing.adoc
simon-dew: Done

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
simon-dew: Reworded as recommended

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
simon-dew: Reworded as suggested.

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
simon-dew: Done

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
simon-dew: Done

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
simon-dew: Reworded "cannot be subclassed"

modules/ROOT/pages/_partials/commons/common-indexing.adoc
simon-dew: Good spot, fixed

modules/ROOT/pages/_partials/commons/common-indexing.adoc
simon-dew: Good spot, fixed

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
simon-dew: Okay, I see it's there

modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc
simon-dew: Done

modules/csharp/pages/new-logging-api.adoc
simon-dew: Okay, I _think_ these have all been dealt with by other fixes. Please take a look at [common-new-logging-api.adoc](https://github.com/couchbase/docs-couchbase-lite/blob/a9a73099daee390d27b788c3044f27d5d168a4f3/modules/ROOT/pages/_partials/commons/common-new-logging-api.adoc) and make sure I haven't missed anything.

modules/ROOT/pages/_partials/commons/common-indexing.adoc
```diff
@@ -109,14 +109,201 @@ include::{root-partials}block_tabbed_code_example.adoc[]
...
+1. If `W` is AND-connected terms, and `X` is OR-connected terms and if any terms of `W` appears as a term of `X`, the partial index is allowed to be used.
+
+2. If a term in `X` is of the form `"z IS NOT MISSING"` and if a term in `W` is a comparison operator on `z` other than `"IS"`, the partial index is allowed to be used. 
+The operators include `=`, `==`, `<`, `>`, `\<=`, `>=`, `<>`, `IN`, `LIKE`, and `BETWEEN`.
+
+IMPORTANT: If `X` is in the form of `"z is NOT NULL"` or `"z is VALUED"`, the first rule must be satisfied.
+
+For example, let the partial index be `c IS NOT NULL` when creating a partial index on collection named `col1` in the default scope.
+
+Then any query that uses operators `=`, `<`, `>`, `\<=`, `>=`, `<>`, `IN`, `LIKE`, or `GLOB` on column `c` would be usable with the partial index, because those comparison operators are only true if `c` is not NULL.
```
pasin: Just noticed this, we don't support GLOB and BETWEEN is missing. So basically, changing BLOB to BETWEEN.

pasin (approved):
Approved with one small change request.

modules/ROOT/pages/_partials/commons/common-indexing.adoc
simon-dew: Done

idulo:
@simon-dew think we should pull the benefits of the new Logging API earlier in the section and communicate it is easy to migrate, encourage earlier migration.

idulo:
@simon-dew on the Partial Index support, we have a section: Partial Full-Text Index, so we call it Full Text, but in practice this is related to the Full Text Search (FTS) feature and it should be with consistent naming. Also we should deep link to the explanation around FTS and the match() function -> e.g. https://docs.couchbase.com/couchbase-lite/current/c/fts.html#use-index as user might not have context for what what we currently call Full Text is and how it can be used.

simon-dew:
> @simon-dew think we should pull the benefits of the new Logging API earlier in the section and communicate it is easy to migrate, encourage earlier migration.

Done

> @simon-dew on the Partial Index support, we have a section: Partial Full-Text Index, so we call it Full Text, but in practice this is related to the Full Text Search (FTS) feature and it should be with consistent naming. Also we should deep link to the explanation around FTS and the match() function.

According to our style guide, the terms should be "Search" (for the service) and "Search index". Mobile is out-of-step on this, but correcting the use of these terms is beyond the scope of this PR. The most common usage in the CBL docs is "full-text index", so I've stuck with "Partial Full-Text Index" for consistency within Mobile. I've added a brief note explaining the `match()` function, and a link to the **Using Full-Text Search — Working with Queries** page.

modules/csharp/pages/new-logging-api.adoc
borrrden: Looks good to me

simon-dew: Fantastic, thanks

PR #953, opened 2025-02-19
Sync iOS and C Partial Index Code Snippets

pasin:
Synced iOS and C Partial Index Code Snippets to be inline with .NET.


ElliotFrancisHunter (approved):
LGTM 👍🏼

PR #955, opened 2025-02-24
Re-add ::set-custom-logging for iOS

velicuvlad:


ElliotFrancisHunter (approved):
Tested and looks like it's fixed! Will integrate now

PR #961, opened 2025-04-01
DOC-13056 simplify gs-* (Getting Started) pages

osfameron:
Through some more exciting regexes like https://github.com/couchbase/docs-couchbase-lite/commit/99e3043ccb7f5afece93d93791f8fb72fdef4034 😱  I've managed to demangle some more bits...

Preview at https://preview.docs-test.couchbase.com/docs-couchbase-lite-DOC-13056-simplify-gs-pages/couchbase-lite/current/android/gs-prereqs.html
(the Getting Started sections for all mobile SDKs)


modules/android/pages/gs-prereqs.adoc
```diff
@@ -1,20 +1,27 @@
...
-ifdef::show_edition[:page-edition: {release}]
-ifdef::prerelease[:page-status: {prerelease}]
 :page-role:
 :description: Prerequisites for the installation of Couchbase Lite
 :keywords: edge nosql api android java
 
-include::partial$_set_page_context_for_android.adoc[]
 
-:param-abstract: Laying out some of the pre-requisites and preparatory steps to be considered before installing Couchbase Lite for {param-name}
-// DO NOT EDIT
```
osfameron, line 12: You have no idea how much I've enjoyed deleting all the "DO NOT DELETE" warnings...

modules/c/pages/gs-downloads.adoc
```diff
@@ -1,60 +1,1207 @@
...
 <<release-3-2-0>>   |
 <<vs-release-1-0-0>>   |
-// | 
+// |
 ****
 
 // This block will always represent the major release version
-:param-version: 3.2.2
-:param-version-hyphenated: 3-2-2
-include::partial$downloadslist.adoc[]
```
osfameron, line 37: I'd be tempted to retain the `downloadslist.adoc` partial myself, as the long lists full of download URLs are easy to make *mistakes in* if we try to edit them manually.

But leaving as is for @simon-dew to look and make a call!

modules/java/pages/gs-prereqs.adoc
```diff
@@ -1,76 +1,83 @@
...
 == Additional Steps For Linux
 
-Before proceeding to xref:gs-install.adoc[], you will need to make the supplied support libraries available to your running application.
+Before proceeding to xref:java:gs-install.adoc[], you will need to make the supplied support libraries available to your running application.
 
+[#steps]
 === Steps
 
-. Download the _zip_ file from here -- https://packages.couchbase.com/releases/couchbase-lite-java/{version-maintenance-java}/couchbase-lite-java-linux-supportlibs-{version-maintenance-java}.zip
+. Download the _zip_ file from here -- https://packages.couchbase.com/releases/couchbase-lite-java/{major}.{minor}.{maintenance-java}{empty}/couchbase-lite-java-linux-supportlibs-{major}.{minor}.{maintenance-java}{empty}.zip
```
osfameron, line 68: Not sure, but it looks like `{empty}` isn't defined by us, and isn't even a builtin asciidoc/Antora attribute?
e.g. it resolves to the empty string, but only accidentally... 🙀 

I suspect it "does no harm" here, but probably worth deleting if it also does nothing useful.

modules/java/pages/gs-install.adoc
```diff
@@ -517,9 +537,52 @@ NOTE:  The generated war file will be at <PROJECT ROOT>/build/libs.
...
+https://docs.couchbase.com/tutorials/[Tutorials]
+
+.
+
+
+++++
+</div>
+++++
 
+:page-toclevels: 2
```
osfameron: I think this is resetting the earlier override (e.g. another artefact of Antora Assembler)

modules/java/pages/gs-prereqs.adoc
osfameron: Ah yes, the pattern is used in the `====== {empty}` hack... 
So we can't just delete it globally, but could try something more precise if Simon thinks worth doing.

modules/android/pages/gs-prereqs.adoc
```diff
@@ -64,4 +72,47 @@ To verify whether your device supports the AVX2 instructions set, https://www.in
...
+
+.
+
+
+[.column]
+====== {empty}
+.Dive Deeper . . .
+https://forums.couchbase.com/c/mobile/14[Mobile Forum] |
+https://blog.couchbase.com/[Blog] |
+https://docs.couchbase.com/tutorials/[Tutorials]
```
TimLFletcher, line 109: Is it worth changing this link to https://developer.couchbase.com/tutorials to avoid the redirect? I notice it seems to add query param to the url, are we using this for tracking purposes?

modules/c/pages/gs-downloads.adoc
```diff
@@ -1,60 +1,1207 @@
...
+xref:c:gs-build.adoc[Build and Run]
 
-:download-path: {url-downloads-mobile}
-:beta-url: https://packages.couchbase.com/releases/couchbase-lite-c/{version-full}/
 
- 
 .Downloads are available for the following versions:
 ****
-<<release-3-2-2>>   |
 <<release-3-2-1>>   |
```
TimLFletcher, line 29: Maybe I'm missing something but we've regressed a version here.

TimLFletcher: (and several other places on this page, the live version lists 3.2.2 whereas the preview is still on 3.2.1)

modules/c/pages/gs-install.adoc
```diff
@@ -487,33 +479,814 @@ In the code, before opening the database and using the vector search extension,
...
+--
+
+[#android]
+=== Android
+
+[cols="^1,^1,^1,^1,^1",options="header"]
+|===
+.>| API | x86 | x64 .>| ARM 32 .>| ARM 64
+
+| 22+ | image:couchbase-lite/current/{underscore}images/yes.png[] | image:couchbase-lite/current/{underscore}images/yes.png[] | image:couchbase-lite/current/{underscore}images/yes.png[] | image:couchbase-lite/current/{underscore}images/yes.png[]
```
TimLFletcher: The yes.png isn't working on the preview for any of these platforms

modules/java/pages/gs-install.adoc
```diff
@@ -42,94 +46,100 @@ Include the following in your Gradle `build.gradle` or Maven `pom.xml` file, as
...
 +
 Windows requires {Cpp} runtime installed.
 Please install the Visual {Cpp} Redistribution package from this link:
 https://www.microsoft.com/en-us/download/details.aspx?id=52685
 
 * Deploying to Linux? +
-You need to deploy the {cbl} `support` library, which is available _only_ on the zip download distributable.
-See the _Additional Steps for Linux_ section in xref:gs-prereqs.adoc#additional-steps-for-linux[].
+You need to deploy the Couchbase{nbsp}Lite `support` library, which is available _only_ on the zip download distributable.
+See the _Additional Steps for Linux_ section in <<additional-steps-for-linux>>.
```
TimLFletcher: This link doesn't seem to work for me.

modules/java/pages/gs-install.adoc
```diff
@@ -421,22 +439,23 @@ image::GradleMenuWebApp.png[,300]
 If you want to deploy your app to a local tomcat container then see <<Deploying a WAR file to tomcat>>
```
TimLFletcher, line 439: This link doesn't work for me in preview or live.

modules/java/pages/gs-build.adoc
```diff
@@ -71,7 +69,7 @@ Web App::
 +
```
TimLFletcher, line 70: The <\<About the Getting Started App\>> link doesn't work for me in preview or live

modules/java/pages/gs-build.adoc
```diff
@@ -135,24 +130,26 @@ About the Starter App::
...
-The {nmMobStarterApp} app will:
+The GettingStarted app will:
 
 * Create a database
 +
-The app creates its database in the `{nmSampleAppDbPath}` directory relative to its root location when run (See: {loc--finding-db-file--xref}).
+The app creates its database in the `/getting-started.cblite2` directory relative to its root location when run (See: <<lbl-find-db-loc,Finding a Database File>>).
 
-TIP: Explicitly state your required database location when creating your database (see: {loc--finding-db-file--xref} for how to do this)
+TIP: Explicitly state your required database location when creating your database (see: <<lbl-find-db-loc,Finding a Database File>> for how to do this)
```
TimLFletcher: This link and the same link above don't work in preview but do work in live.

modules/swift/pages/gs-install.adoc
```diff
@@ -30,35 +37,241 @@ Couchbase Lite is distributed as an XCFramework, to support this:
...
 --
+
 Swift Package Manager::
 +
 --
 NOTE: Using Swift Package Manager to install CouchbaseLiteSwift requires XCode 12+
 
-include::partial$gs-install-tab-spm.adoc[]
+This tab explains how to include the CouchbaseLiteSwift package within your app +
+See: <<case-1>> | <<case-2>>
```
TimLFletcher, line 144: These links do not work for me.

modules/swift/pages/gs-install.adoc
```diff
@@ -144,42 +360,40 @@ end
...
 --
 NOTE: Using Swift Package Manager to install CouchbaseLiteVectorSearch requires XCode 12+
 
-:url-ee: https://github.com/couchbase/couchbase-lite-vector-search-spm.git
-:saved-caption: {example-caption}
-:example-caption!:
-// :this-release: {version-maintenance}
+// :this-release: {major}.{minor}.{maintenance-ios}{empty}
 This tab explains how to include the CouchbaseLiteVectorSearch package within your app +
 See: <<case-1>> | <<case-2>>
```
TimLFletcher, line 372: Same issue here - these links don't work for me.

TimLFletcher:
I've finished my review. No major breakages. Some links don't work, one doc seems to be a version behind, and there's an image or two that seem to be missing. I've left comments.

osfameron:
Thanks @TimLFletcher - I don't see the comments though, are they still left as Pending perhaps?

modules/android/pages/gs-prereqs.adoc
osfameron: yes, the `_gl=` parameter seems to be a Google Analytics thing

modules/c/pages/gs-downloads.adoc
osfameron: Good catch, probably need to update my sources (😱 🍝)

modules/c/pages/gs-install.adoc
osfameron: `%7Bunderscore%7Dimages/yes.png` huh.
I guess the variable definition is lost somewhere 🤔

modules/java/pages/gs-install.adoc
osfameron: Should be `<<Deploying a WAR file>>` I think.
I've avoided making *fixes* to live errors in the automated extraction part 🤔 
will add a list of manual steps to do in an additional commit.

modules/swift/pages/gs-install.adoc
osfameron: @simon-dew noticed that tabbed sections weren't working, possibly due to having auto-generated `[#anchor]` artefacts from Antora Assembler.

As an "optimization" I simply deleted them within tab blocks...
Looks like I need to be more careful with which ones to delete... 🤦

osfameron: OK that issue was https://github.com/couchbase/docs-couchbase-lite/pull/956#discussion_r2014162434

A *generated* link *before* a dlist header will break the tabs.
This is a *manual* link *inside* the dlist header, so doesn't.

I don't understand why in this case the generated one doesn't happen, but it seems like reverting the "fix" (necessary for that p2p sync PR) works here 😕 🤯

modules/android/pages/gs-prereqs.adoc
osfameron: The only reason to NOT make this change would be if we think we might ever take Tutorials **back** from DA, or move them to another location?

Having the link go to us first means that we have control over the ultimate location, otherwise we rely on them to change it.

Don't mind either way, let me know!

modules/java/pages/gs-build.adoc
osfameron: Fixed with:

```
# <<java:database:::lbl-find-db-loc,Finding a Database File>>
s/<<(.*?):::([a-zA-Z-]+)(,(.*?))?>>/xref:$1.adoc#$2\[$4]/g;
```

This is FINE 🔥

osfameron:
Refreshed the preview, should have fixed all the issues except:
* ones also in Live
* updating that tutorials URL

modules/android/pages/gs-prereqs.adoc
TimLFletcher: Fine to leave it for now. Let's focus.

TimLFletcher (approved):
Looks good to me 👍

PR #966, opened 2025-04-29
DOC-13226 simplify remaining "single" pages

osfameron:
e.g. the common pages that don't have expandable nav:

* Scopes and Collections
* Documents
* Blobs
* Indexing
* Handling Data Conflicts
* Upgrade

Preview at https://preview.docs-test.couchbase.com/docs-couchbase-lite-DOC-13226-single-pages/ 

Features in this iteration: demangling `pass:q,a[ ... ]` inclusions... 😱


osfameron:
~~Weird artefact in "see also" links:~~

*update* fixed by subsequent commit, ready for review.

modules/android/pages/document.adoc
```diff
@@ -1,11 +1,1270 @@
...
+
+The document body is stored in an internal, efficient, binary form called
+https://github.com/couchbaselabs/fleece#readme[Fleece].
+This internal form can be easily converted into a manageable native dictionary format for manipulation in applications.
+
+Fleece data is stored in the smallest format that will hold the value whilst maintaining the integrity of the value.
+
+
+[#fleece-data-encoding]
+== Fleece data encoding
```
TimLFletcher: I think this ought to be === ?

modules/android/pages/indexing.adoc
```diff
@@ -1,19 +1,390 @@
+
```
TimLFletcher, line 1: Looks like we're a version behind, here. The live version is different - updated for 3.2.2.

modules/android/pages/upgrade.adoc
```diff
@@ -1,13 +1,429 @@
+
```
TimLFletcher, line 1: Again, we're a version behind here. 3.2.2 vs 3.2.3

TimLFletcher (commented):
I did a diligent review of the android and C pages, then a quicker review of the remaining sections.

I didn't see anything broken in those latter sections BUT the comments from the first sections pertain to them - namely that I think we're behind in some versions.

modules/android/pages/document.adoc
osfameron: Gah - this one is annoying... because Collector flattens everything into a single page, the heading levels get mangled based on both the *page* nav *and* the heading ordering... 

So I try to fix that with a heuristic (which is, as they say, a fancy way to say "doesn't work" 😁).
There might be a few more of those lurking around unfortunately.

modules/android/pages/indexing.adoc
osfameron: I have updated the Rube Goldberg machine... 😓

PR #975, opened 2025-05-21
DOC-13261 simplify query pages

osfameron:
https://preview.docs-test.couchbase.com/docs-couchbase-lite-DOC-13261-simplify-query-pages/


osfameron:
slight mangling of the Live / Troubleshooting pages, at least for C.
image

... ah yes, it's Embargoed, so shouldn't be in the nav (and because of assumptions in the simplification script, garbles the previous and following articles too...)

... fixed

TimLFletcher (approved):
Blimey that's a big one. I gave them the "is anything obviously broken or missing?" check. Looks all good.

Why did we remove the C query troubleshooting page? I assume that's intentional.

osfameron:
> Why did we remove the C query troubleshooting page? I assume that's intentional.
yup, it was broken on live, and set to Embargoed (e.g. shouldn't have appeared on the Nav at all)

PR #977, opened 2025-05-30
DOC-13273 mopup of partials, including "the commons"

osfameron:
Finally attempting to kill off the "commons":

This is mostly deletions, but also tidyups, a few extra "simplification" runs, and other fixups.

Preview at:

https://preview.docs-test.couchbase.com/docs-couchbase-lite-DOC-13273-mopup-common/ 

CC'ing @RichardSmedley this time, as he'll be forking from 3.2 to start work on 3.3.


RichardSmedley (approved):
I don't understand all of it, but it looks like a big step in the right direction overall.

PR #979, opened 2025-06-06
DOC-13296 fixup gs downloads

osfameron:
~Draft as missing info on 3 broken links (asked in #mobile-docs-related) though this fix could be merged in separately if needed.~

Preview building to https://preview.docs-test.couchbase.com/docs-couchbase-lite-DOC-13296-fixup-gs-downloads/couchbase-lite/current/c/gs-downloads.html

@RichardSmedley will need porting to 3.3 also.

### DIFF
--- a/modules/c/pages/gs-downloads.adoc
+++ b/modules/c/pages/gs-downloads.adoc
@@ -1662,9 +1662,9 @@ Enterprise::
 
 .4+| Linux
 
-| https://packages.couchbase.com/releases/couchbase-lite-vector-search/1.0.0/couchbase-lite-vector-search-1.0.0-linux-aarch64.zip[couchbase-lite-vector-search-1.0.0-linux-aarch64.zip]
-| https://packages.couchbase.com/releases/couchbase-lite-vector-search/1.0.0/couchbase-lite-vector-search-1.0.0-linux-aarch64.zip.sha256[couchbase-lite-vector-search-1.0.0-linux-aarch64.zip.sha256]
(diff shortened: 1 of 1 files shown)


iamfortune:
LGTM

modules/c/pages/gs-downloads.adoc
simon-dew, line 1: I think the **Community Edition** and **Enterprise Edition** tabs are not working because of the anchors embedded within the tabbed interface blocks, e.g. at line 1581 (and throughout). Not sure if you want to do anything about this. I would remove the anchors within the tabbed interface blocks, and probably the anchors for each "Download link table" heading as well? They're repetitive, and I don't think any of these are needed because the headings already have anchors.

Secondly, in the output, each tabbed interface block is captioned "Example:" when you really want it to be captioned as a table. Again, not sure if you want to do anything about this. Could hide example captions for this page? Or just remove the "Download link table" headings, because again I'm not sure that they're needed.

simon-dew (commented):
Checked all the updated links and they are fine 👍 

However I noticed a couple of unrelated issues with the markup of the page

modules/c/pages/gs-downloads.adoc
osfameron: @simon-dew I was doing the minimal to fix the reported issue, but I could look over this with Fortune later this week as a learning exercise, if you think that's appropriate.

simon-dew: Happy to +1 if you need a quick fix! I'll make a note of these issues and raise a ticket later

simon-dew (approved):
LGTM 👍 — the updated links all work

PR #1026, opened 2025-10-03
[DOC-13517]: Feedback on Live Query—Working with Queries | Couchbase Docs

RayOffiah:
Updated the Live Query documentation to correct a broken link leading to the Kotlin example.


RayOffiah:
Branch name contains hidden characters.

PR #1087, opened 2026-03-31
Fix PR Validation

pasin:
* iOS : 
   - Fixed and simplify script to download the right CBL and VS binary.
   - Fixed one error in code snippet.
  
* C : 
   - Applied the same method as iOS get the lastest CBL version.
   - Removed obsolete logging code snippet as the API has been changed in 4.0
  
*  Android:
   - Updated AGP from 8.0.2 to 8.13.2 to fix JDK 21 incompatibility issue.
   - Update Android script to get CBL and VS version based on isReleased flag.


iamfortune (approved):
Lgtm

PR #1110, opened 2026-06-26
Update net

iamfortune:
Update the .NET platform

### DIFF
--- a/modules/csharp/pages/supported-os.adoc
+++ b/modules/csharp/pages/supported-os.adoc
@@ -18,8 +18,11 @@ Related Content -- xref:cbl-whatsnew.adoc[What's New]  |  xref:csharp:releasenot
 
 The following table identifies the supported platforms.
 
-Run-times which have received more testing and are *officially* supported are shown in xref:csharp:supported-os.adoc#supported-os-versions[]:
+Run-times which have received more testing and are *officially* supported are shown in xref:csharp:supported-os.adoc#supported-os-versions[].
 
+[NOTE]
+.Newer .NET Runtime Versions
+--
 Later versions, including .NET 11 and higher, also work because .NET provides backward compatibility.
 Couchbase Lite is built to be compatible with newer .NET runtime versions as they're released.
 If you encounter any issues with a newer .NET version, submit a support ticket.
@@ -29,7 +32,7 @@ If you encounter any issues with a newer .NET version, submit a support ticket.
 [#supported-os-versions]
 [#supported-os-versions,cols="1,^2,^2", options="header"]
 |===
-.>|.NET Runtime .>|Minimum Runtime Version (and newer).>|Minimum OS version
+.>|.NET Runtime .>|Minimum Runtime Version (and newer) .>|Minimum OS version
 
 | .NET Framework
 | 4.6.2
@@ -57,7 +60,6 @@ a| MacOS 14
 |10.0
 |API 24
 
-
 |===
 
 
@@ -75,7 +77,8 @@ The following run-times are compatible but are not QE tested, and so are not off
 
 |.NET Linux
 | 8.0
-|n/a*
+|n/a
+
 |===
 
 
--- a/modules/csharp/pages/supported-os.adoc
+++ b/modules/csharp/pages/supported-os.adoc
@@ -103,7 +103,7 @@ They are tested on Ubuntu 20.04, but have been shown to work on CentOS, and in t
 .>|.NET Runtime .>|Version Removed .>|Deprecation Release
 
 |Windows 10
-|3.4 / 4.1
+|4.1
 |4.0
 
 |===
--- a/modules/csharp/pages/supported-os.adoc
+++ b/modules/csharp/pages/supported-os.adoc
@@ -20,11 +20,16 @@ The following table identifies the supported platforms.
 
 Run-times which have received more testing and are *officially* supported are shown in xref:csharp:supported-os.adoc#supported-os-versions[]:
 
+Later versions, including .NET 11 and higher, also work because .NET provides backward compatibility.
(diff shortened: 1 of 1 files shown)


modules/csharp/pages/supported-os.adoc
```diff
@@ -66,7 +62,15 @@ a| MacOS 13
 
 |.NET Android
 |10.0
-|API 24
+|API 22+
```
borrrden: This has gone backwards.  .NET Android was already correct and now it's gone back to its 3.2 version

modules/csharp/pages/supported-os.adoc
```diff
@@ -66,7 +62,15 @@ a| MacOS 13
...
+
+|Xamarin Android
+|10+
+|API 22
+
+|Xamarin iOS
+|10+
+|10
```
borrrden: Don't re-add Xamarin

modules/csharp/pages/supported-os.adoc
```diff
@@ -81,10 +85,25 @@ The following run-times are compatible but are not QE tested, and so are not off
...
+|14
 
 |.NET Linux
 | 8.0
-|n/a
+|n/a*
+|===
+
+*{sp}There are many different variants of Linux, and we don't have the resources to test all of them.
+They are tested on Ubuntu 20.04, but have been shown to work on CentOS, and in theory work on any distro supported by .NET.
```
borrrden: Remove the "they are tested" part, because they are not

modules/csharp/pages/supported-os.adoc
```diff
@@ -23,38 +23,34 @@ Run-times which have received more testing and are *officially* supported are sh
...
---
-The *Minimum Runtime Version* column specifies the minimum required version for each .NET runtime.
-Later versions, including .NET 11 and higher, also work because .NET provides backward compatibility.
-Couchbase Lite is built to be compatible with newer .NET runtime versions as they're released.
-If you encounter any issues with a newer .NET version, submit a support ticket.
+* Xamarin Android - All Versions
+* Xamarin iOS - All Versions
+* .NET Desktop - 6
+
+Please plan to migrate your apps to use an appropriate alternative version.
```
borrrden: This is also pulled out of 3.2 over the more recent 4.0 notes.  How does it end up like this?

borrrden (changes requested):

modules/csharp/pages/supported-os.adoc
```diff
@@ -20,41 +20,25 @@ The following table identifies the supported platforms.
...
---
-
-[NOTE]
-.Newer .NET Runtime Versions
---
-The *Minimum Runtime Version* column specifies the minimum required version for each .NET runtime.
-Later versions, including .NET 11 and higher, also work because .NET provides backward compatibility.
-Couchbase Lite is built to be compatible with newer .NET runtime versions as they're released.
-If you encounter any issues with a newer .NET version, submit a support ticket.
---
```
borrrden, line 33: This NOTE should be kept because it continues to be important for users who get confused about what we actually support

modules/csharp/pages/supported-os.adoc
```diff
@@ -20,41 +20,25 @@ The following table identifies the supported platforms.
...
-
 .Supported versions
 [#supported-os-versions]
 [#supported-os-versions,cols="1,^2,^2", options="header"]
 |===
-.>|.NET Runtime .>|Minimum Runtime Version (and newer) .>|Minimum OS version
```
borrrden: This language was specifically added because of repeated confusion.  Do you remember that conversation?

modules/csharp/pages/supported-os.adoc
```diff
@@ -81,10 +66,25 @@ The following run-times are compatible but are not QE tested, and so are not off
...
+|14
 
 |.NET Linux
 | 8.0
-|n/a
+|n/a*
+|===
+
+*{sp}There are many different variants of Linux, and testing resources are limited.
+Couchbase Lite has been shown to work on CentOS, and in theory it works on any distro supported by .NET.
```
borrrden: This is no longer true.  We don't test on CentOS because CentOS is dead.  I think this note should just be dropped at this point.

borrrden (dismissed):

modules/csharp/pages/supported-os.adoc
iamfortune: I've taken it out

modules/csharp/pages/supported-os.adoc
```diff
@@ -20,17 +20,6 @@ The following table identifies the supported platforms.
...
-.Deprecation Notice
---
-Couchbase Lite 4.0 deprecates Windows 10 support and will remove it in a future release.
-Plan to migrate your apps to use an appropriate alternative version.
---
-
-[NOTE]
-.Newer .NET Runtime Versions
---
-The *Minimum Runtime Version* column specifies the minimum required version for each .NET runtime.
```
borrrden: The first sentence is still deleted

modules/csharp/pages/supported-os.adoc
```diff
@@ -81,10 +71,23 @@ The following run-times are compatible but are not QE tested, and so are not off
 .>|.NET Runtime .>|Minimum Runtime Version .>|Minimum OS version
 |.NET Mac
 | 8.0
-|13
+|14
 
 |.NET Linux
 | 8.0
-|n/a
```
borrrden: Not sure if you want to keep the `*` or not

idulo (approved):
Approval contingent on the comments above addressed in a fast follow PR

modules/csharp/pages/supported-os.adoc
borrrden, line 84: Still recommend removing this since we do no testing whatsoever.

PR #1130, opened 2026-07-29
Callout-list fixes

osfameron:
Noticed we had some build errors with "Phrase not found", that were related to our use of @RayOffiah's plugin for external callouts.

Using Claude, it identified that this was introduced by the refactor to tabs. I've verified that the errors seem to go away, and some callouts now display.

Here's Claude's summary (includes a few suggestions for ongoing fixes):

> Callout-list relocation (fixes "Phrase not found") — moved 6 @/.../ callout lists from after the [tabs] block back to inside the C tab (immediately after the C include::, before that tab closes), so asciidoctor-external-callout binds to the correct C listing instead of the trailing C++ one:
> - p2psync-websocket.adoc: listener-simple and replicator-simple blocks
> - p2psync-websocket-using-passive.adoc: listener-initialize, listener-config-tls-enable;...tls-id-anon, listener-config-client-auth-root, listener-config-client-auth-lambda blocks
> 
> (p2psync-websocket-using-active.adoc needed no relocation — its one @/.../ list sits on a plain literal code block, not a tabbed one.)
> 
> Wrong include path: fixed code_snippets/cbl_cpp.cpp → code_snippets_cpp/cbl_cpp.cpp everywhere across
> all three pages (19 occurrences total: 1 in websocket.adoc, 6 in active, 13 in passive).
> 
> Heads up, unresolved: fixing the path doesn't make the Passive Peer page's C++ tabs actually work — code_snippets_cpp/cbl_cpp.cpp has zero listener-* tags. The C++ (cbl:: API) example code for the Listener/server side hasn't been written yet, so every {cpp}/C++ tab in p2psync-websocket-using-passive.adoc (and the listener-simple C++ tab in p2psync-websocket.adoc) will still fail to resolve — just with an Antora "tag not found" error instead of "file not found." That's a content gap for whoever owns the C++ snippets, not a doc-syntax fix. Worth a ticket if there isn't one already (the client-auth-root block even had a comment referencing a prior cleanup ticket for the same file, so there's precedent for tracking it that way).
> 
> Also noticed while sweeping: conflict.adoc, query-resultsets.adoc, query-live.adoc, and query-n1ql-mobile.adoc have the same wrong code_snippets/cbl_cpp.cpp path — outside the three pages you asked about, so I left those alone, but flagging in case they're on your list too.

### DIFF
--- a/modules/c/pages/p2psync-websocket-using-active.adoc
+++ b/modules/c/pages/p2psync-websocket-using-active.adoc
@@ -335,7 +335,7 @@ include::c:example$code_snippets_cpp/cbl_cpp.cpp[tags=replication-retry-config,i
 --
 [source, cpp]
 ----
-include::c:example$code_snippets/cbl_cpp.cpp[tags="replication-retry-config", indent=0]
+include::c:example$code_snippets_cpp/cbl_cpp.cpp[tags="replication-retry-config", indent=0]
 ----
 <.> Here we use heartbeat to set the required interval (in seconds) between the heartbeat pulses.
 <.> Here we use maxAttempts to set the required number of retry attempts.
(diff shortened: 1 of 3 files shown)


iamfortune (approved):
this lgtm

PR #1133, opened 2026-09-08
DOC-14677: Fix broken xrefs and includes in couchbase-lite 4.1

osfameron:
This PR fixes 20 of 169 build errors on couchbase-lite  4.1 — 
all mechanical (missing module prefixes, a wrong directory name, and a dead include left over from the DOC-13273 partial-flattening work). 

## Remaining 149 errors 🙈 

The remaining 149 fall into a few concentrated buckets, two of which are large content/architecture decisions rather than link fixes:

  1. "Peer-to-Peer Sync" landing page was deliberately un-published (~45 errors)
  landing-p2psync.adoc is linked from dozens of pages across every platform. It wasn't deleted — it was
  renamed with a leading dot (.landing-p2psync.adoc), Antora's convention for hiding a page while keeping its
  content in the repo. Only the C module still has this hidden file with real content; every other platform's
  copy is gone entirely. Needs a decision: republish, redirect these links elsewhere
  (landing-replications.adoc exists on some platforms), or confirm this was intentional and strip the links.

  2. JavaScript rows in the six compatibility.adoc tables assume JS is a module inside couchbase-lite (~42
  errors) It isn't — modules/javascript/ was migrated out into its own top-level repo/component, couchbase-lite-js
  (currently versioned 1.0, not tracking the main  product's 2.8–4.1 numbering). There's no equivalent
  per-version "supported OS" page over there for the historical-version comparison row to link to. Real fix
  requires deciding what (if anything) that row should show now.

  3. csharp:gs-prereqs.adoc (14 errors) — page doesn't exist; gs-install.adoc looks like the likely successor
  but has no "Prerequisites" section to anchor into.

  4. c:querybuilder.adoc (6 errors) — every other platform has QueryBuilder docs; C does not. These look
  like "Related Content" footers copy-pasted from another platform's template without adjusting for C, where the
  API doesn't apply.

  5. swift/java/csharp/android:supported-os.adoc pinned to old versions 2.8–4.0 (24 errors) — pages genuinely
  exist at those historical branches, so this looks like a playbook-scope issue (same pattern as the SDK version
  pins we found in docs-server), not a content bug.

  6. A handful of one-offs needing individual triage: objc:new-logging-logs.adoc (likely renamed to
  new-logging-api.adoc), objc:replication.adoc linking to swift:index.adoc (wrong platform),
  csharp:replication:p2psync-websocket.adoc (module/family mismatch), kotlin:supported-os.adoc,
  android:ppsync-multipeer-transports.adoc.

### DIFF
--- a/modules/android/pages/_partials/_set_page_context_for_android.adoc
+++ b/modules/android/pages/_partials/_set_page_context_for_android.adoc
@@ -8,8 +8,6 @@ ifdef::is_diag[_set_page_context_for_android]
 :param_is_root!:
 :version-maintenance-android: {major}.{minor}.{maintenance-android}
 
-include::ROOT:partial$_set_page_context.adoc[]
-
 // BEGIN::module page attributes
 // :module: {param-module}
 // :packageNm: couchbase-lite-{module}
--- a/modules/android/pages/compatibility.adoc
+++ b/modules/android/pages/compatibility.adoc
@@ -246,7 +246,7 @@ h| .NET
 * xref:android:releasenotes.adoc[Release Notes]
 * xref:android:compatibility.adoc[Compatibility]
 * xref:android:supported-os.adoc[Supported Platforms]
-* xref:cbl-whatsnew.adoc[What's New]
+* xref:ROOT:cbl-whatsnew.adoc[What's New]
 
 .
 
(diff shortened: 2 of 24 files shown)


iamfortune (approved):
lgtm

